PR #22, opened 2022-03-20
Data Flow setup intro, diagram, schema create section

swcurran:

### DIFF
--- a/spec/data_flow_setup.md
+++ b/spec/data_flow_setup.md
@@ -6,7 +6,7 @@ This section of the specification describes the major data flows within AnonCred
 * Credential issuance, including operations by both the Issuer and Holder
 * Present Proof, including operations by both the Verifier and the Holder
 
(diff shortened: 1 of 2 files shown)


swcurran:
Adds content to the Data Flow: Setup section of the specification, including an intro to the whole section and the create schema subsection.

A plantuml diagram is also included but not referenced. Instead, I used the inline mermaid format. Once we see the rendered mermaid diagram in the published spec, we'll decided if it is "good enough", and if so remove the plantuml diagram from the spec/ folder. If not, we'll likely include both the plantuml and rendered diagram, and reference the rendered diagram in the spec.

matt-raffel-kiva (approved):

swcurran:
Looks like there is a bug in the mermaid diagram. I think I've found the source and will have a new PR to fix just that...

Sigh!

PR #36, opened 2022-04-08
Pass at defining the CredDef structure more thoroughly

wip-abramson:
This is still probably WIP. Not sure if it is too much in the math in parts.

I based the math off https://github.com/hyperledger-archives/indy-anoncreds/blob/master/docs/anoncred-usecase1.pdf and the implementation in Ursa. They seem to match.


swcurran (changes requested):
I think this looks great. 

If you can (although not the end of the world) please remove the docs/index.html from the PR. I've got it in the .gitignore, but that doesn't seem to be working. It's not needed in the main branch and is force pushed to the ghpages branch for publication.

I think the level of detail is right.  I would suggest adding to the "References" section an entry for what you used as source material, and include that is what you did in the PR.

We have a couple of other documents that need to go into the reference section as well, that I'll try to get in as well.

Great work!

swcurran:
@wip-abramson -- there are DCO issues with this PR and chances are my PR will create lots of merge conflicts. Perhaps it would be best to do a copy/paste and resubmit after my PR is merged -- hopefully by Monday.

PR #39, opened 2022-04-26
WIP: 17 Create the draft data flow issuance document

ArPhil:


ArPhil:
@swcurran , @victormartinez-work : Started working on the mermaid diagram for credential issuance. @hkny and me had a discussion, whether we should include the `propose-credential` steps as defined [here](https://github.com/hyperledger/aries-rfcs/blob/main/features/0036-issue-credential/README.md). 

I'd like to skip the `propose-credential` steps, since it rather belongs to the Aries procotol than to the (Indy) AnonCreds Spec IMO. 

Do you agree?

MichScha:
> In anoncreds, the "offer" step is not optional AFAIK. As I recall, the step is required because the issuer is providing a cryptographic commitment to the holder. The wording you have appears to be from the Aries RFCs (0036, 0453). In that case, because an issuer may be using other than AnonCreds, the step is optional, However, with AnonCreds, it is a required step.
> 
> Those with more knowledge of AnonCreds should weigh in here to check my understanding. And of course, when we get into the details of the credential offer data model, we will have a full explanation included.
> 
> Note: If I'm wrong about the requirement of a Credential Offer in AnonCreds, I think we would try to drop it entirely if it is not part of the cryptographic flow.

By looking into the ursa [code ](https://github.com/hyperledger/ursa/blob/main/libursa/src/cl/issuer.rs#L189) I can´t find a necessary element that has a relation to a previous step except for the `nonce` that is also required in the sign function but this is not shared with a credential offer. The offer has just three elements (`schema_id or cred_def.schema_id, cred_def, key_proof.raw_value`) as per implementation of [acapy](https://github.com/hyperledger/aries-cloudagent-python/blob/4240fa9b192ea4cdb4026211ea4bec694aec5506/aries_cloudagent/indy/credx/issuer.py#L212).
The nonce itself is been shared from the prover via credential_request (`prover_did, cred_def_id, blinded_ms, blinded_ms_correctness_proof, nonce`). Therefor I can´t find a necessary step where we are needing the credential offer - am I missing something?

swcurran (approved):

PR #42, opened 2022-05-08
Add editors and contributors

swcurran:
- Update Discord references to be an invitation
- Add editors and contributors


TelegramSam (approved):
Discussed WG 20220613

PR #44, opened 2022-05-09
Update CredDef data structure with cryptographic details

wip-abramson:


wip-abramson:
So, I am not exactly sure why it is saying DCO missing. I used the -S flag with each of my commits. And since have done the rebasing that they the DCO details suggest ...

swcurran:
@wip-abramson -- could you take a look at fixing the DCO?  See my earlier note about what I think is the issue.

PR #46, opened 2022-05-14
[WIP] First draft of document describing the crypto protocols

ale-linux:
I've created this first draft to show how we might want to describe the crypto protocols. I collected a few questions to discuss with the rest of the team and decide how to proceed:

- should we add references to the code? For example, the algorithm I described is implemented by `_init_eq_proof` in ursa's `libursa/src/cl/prover.rs`.
- these protocols are typically orchestrated in three phases given that they are essentially instantiations of Sigma protocols. Should we also break them down in three phases? Or should we simplify them and assume that each algorithm accumulates t-values and constants in some data structure, that at some point it can invoke a `getC` function and compute s-values?
- some of the random values used to compute t-values need to be reused by various protocols. Should we assume some global context where they are stored, accessible to all protocols?
- the cryptographic context (`context` or `m2` in the code) is treated separately (e.g. it has its own element `R_ctx`). Should we also treat it specially? What about the user secret key?
- should we clarify which formats the inputs should be in?
- should the relationships of the lengths be explained/justified? Some come externally, some have to be computed


swcurran:
I think this is excellent.

My thought is we should include this in a similar way to the the [Terminology](https://anoncreds-wg.github.io/anoncreds-spec/#terminology) section so that we can reference the algorithms in this form`[[ref: ProveCL]]`.  Basically, add a "Cryptographic Protocols" section with `[[def: Name_of_Protocol]]` section. My guess is that it should be towards the end.

swcurran:
@wip-abramson can you please take a look at this one?

spec/crypto.md
```diff
@@ -0,0 +1,140 @@
...
+
+ - TBD
+
+Outputs:
+
+ - A', term of the rerandomised signature
+ - Z~. t-value for the signature
+ - e^, s-value for the signature
+ - v^, s-value for the signature
+ - (m^_1,..., m^_L), s-value for the signature
```
wip-abramson, line 121: Are we saying that pi is some representation of all of these values? (line 93)

wip-abramson: Ah I see we are from line 137

spec/crypto.md
```diff
@@ -0,0 +1,140 @@
...
+
+`ProveCLCommit` finds has its reference implementation [here](https://github.com/hyperledger/ursa/blob/ece6ce32a59df4e1f99fa38243c1236423066bc2/libursa/src/cl/prover.rs#L1313-L1394).
+
+```
+( A', v', e', Z~, e~, v~ ) = ProveCLCommit( PK, signature, (m_1,..., m_L), RevealedIndexes, R )
+
+Inputs:
+
+ - PK (REQUIRED), the public key of the issuer
+ - signature (REQUIRED), a signature generated by the issuer
```
wip-abramson, line 27: "a signature generated by the issuer" I would be tempted to be more specific, because this signature is actually generated by both issuer and holder. The unblinded version of the signature generated by the issuer perhaps?

Its just the protocol that describes this unblinding process will need to refer to the signature generated by the issuer and that signature won't be the same one as here.

wip-abramson:
This looks great to me overall!

Is there a list of all the other cryptographic protocols that need to be defined somewhere? This ProveCl is just one of the proofs that need creating when producing a presentation right?

swcurran:
There is the proof of non-revocation that needs to be defined as part of the presentation generation process.

spec/crypto.md
swcurran: @ale-linux -- this is the only suggested change to this section. Could you please respond and we'll get this merged.

Thanks!

spec/crypto.md
```diff
@@ -0,0 +1,140 @@
+## Cryptographic protocols
+
+### Terminology
+
+_TODO: general terms here_
+
+### Protocols
+
+Most protocols described here are zero-knowledge proofs for different statements. They follow the patterns of three-move rounds protocols with a commit, a challenge and a response phase which we shall describe as three separate algorithms. The challenge phase is unique, whereas the commit and response phases have a specific instance for each of the statements being proved.
```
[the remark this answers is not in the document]
ale-linux: Good idea to provide some link. I'm just not sure I'd pick that article :)

How about these alternatives:

- The [original work](https://d-nb.info/1156214580/34) by Schnorr
- [Cramer's thesis](https://ir.cwi.nl/pub/21438) that first defines the concept of Sigma protocol
- this nice set of [lecture notes](http://www.daimi.au.dk/~ivan/Sigma.pdf) on Sigma protocols
- this [book](https://link.springer.com/book/10.1007/978-3-642-14303-8) with a good section on the subject.

Wdyt?

spec/crypto.md
ale-linux: Good point - how about "the output of the issuance protocol"?

ale-linux:
> This looks great to me overall!
> 
> Is there a list of all the other cryptographic protocols that need to be defined somewhere? This ProveCl is just one of the proofs that need creating when producing a presentation right?

The idemix technical report contains a list, but I'm not sure whether it's complete - the only way to make sure is to review the ursa code which ultimately is the ground truth. That's what I'm doing right now

ale-linux:
> There is the proof of non-revocation that needs to be defined as part of the presentation generation process.

Yup - that's an additional protocol - do you think we should treat it here as part of this? It can be decoupled nicely - but in the code it's actually part of the same function. Thoughts?

swcurran:
Definitely we shouldn't do it in this PR, and I would decouple it from this part -- a separate section in the Crypto part of the spec. I'm working on the high level overview of the revocation.

TelegramSam:
Discussed WG 20220627. @ale-linux to add some revisions.

spec/crypto.md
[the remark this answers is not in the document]
ale-linux: Added a subsection in the informative references page

spec/crypto.md
ale-linux: Done

ale-linux:
Done @TelegramSam, thanks!

ale-linux:
I think this is ready to merge right @swcurran?

swcurran (approved):

PR #49, opened 2022-06-26
First cut at the revocation section

swcurran:

### DIFF
--- a/spec/data_flow_revocation.md
+++ b/spec/data_flow_revocation.md
@@ -1,11 +1,430 @@
 ### AnonCreds Revocation Data Flow
 
-::: todo
-Sumamrize the impacts of revocation on the various other flows
-:::
+AnonCreds includes a mechanism that supports the revocation of verifiable
+credentials. This mechanism includes:
+
+- An [[ref: issuer]] setting up to issue revocable credentials.
+- An [[ref: issuer]] issuing revocable credentials.
+- An [[ref: issuer]] revoking issued credentials.
+- A [[ref: verifier]] requesting a presentation to include a proof of non-revocation
+  for one or more revocable credentials.
+- A [[ref: holder]] generating based on the request of the verifier a proof of
+  non-revocation for attributes derived from revocable credentials.
+- A [[ref: verifier]] verifying a proof of non-revocation included in a
+  presentation from a [[ref: holder]].
+
+A fundamental goal of AnonCreds is to not provide a correlatable identifier for
+a either a [[ref: holder]] or a credential as part of generation and verification of an AnonCreds
+presentation. Applying that goal to revocation means that the revocation
+mechanism must support the [[ref: holder]] proving a credential used in generating a
+presentation is not revoked without providing a correlatable identifier for that
+credential. As such, the AnonCreds revocation mechanism uses a Zero Knowledge
+Proof (ZKP) that allows the [[ref: holder]] to prove a credential they hold is not
+revoked without revealing an identifier for their credential.
 
 #### AnonCreds Issuer Setup With Revocation
 
+The details of [[ref: issuer]] setting up revokable credential types are covered
+in the [issuer
+setup](data_flow_setup.md#generating-a-creddef-with-revocation-support) section
+of this specification.
+
 #### AnonCreds Issuance with Revocation
 
+The details of an [[ref: issuer]] issuing a revokable credential to a [[ref:
+holder]] are covered in the [issuance data
+flow](data_flow_issuance.md#issue-credential) section of this specification.
+
+If the [[ref: issuer]] has created the [[ref: RevReg]] from which they are
+issuing credentials to have an `ISSUANCE_TYPE` of `ISSUE_ON_DEMAND`, the
+[[ref:issuer]] must publish a [[ref: RevRegEntry]] (as described in the next
+section) as each credential issued. Based on the experience of the AnonCreds
+community in the use of revocable credentials, it is highly recommended the
+`ISSUE_ON_DEMAND` approach **NOT** be used unless absolutely required by your
+use case. The reason for this recommendation is that `ISSUE_ON_DEMAND` requires
+near real-time activation, meaning the issuing of each credential requires the
+publication of a new [[ref: RevRegEntry]], resulting in many [[ref:
+RevRegEntry]] transactions being performed, one per credential issued. As such,
+we would like to deprecate or even remove the `ISSUE_ON_DEMAND` approach in the
+next version of AnonCreds revocation. Feedback from the community on this would
+be appreciated. We are particularly interested in understanding what use cases
+there are for `ISSUE_ON_DEMAND`.
+
+#### AnonCreds Credential Revocation and Publication
+
+When an [[ref: issuer]] decides to revoke a previously issued credential (or
+activate a previously inactive/revoked credential), they do so by publishing
+another instance of the [[ref: RevRegEntry]] object. Recall from the issuer
+setup section, the specification about [creating and publishing the first [[ref:
+RevRegEntry]]](data_flow_setup.md#creating-the-initial-revocation-registry-entry-object)
+for a [[ref: RevReg]]. In that process, the accumulator for the initial state of
+the [[ref: RevReg]] is published. When subsequent [[ref: RevRegEntry]]
+transactions are published to the ledger, each includes an updated value of the
+accumulator, along with lists of the identifiers of the credentials that have
+had their revocation state toggled -- either changed to issued, or changed to
+revoked -- since the last [[ref: RevRegEntry]] was published. An example of the
+data in the [[ref: RevRegEntry]] is shown in the following example of a [[ref:
+RevRegEntry]], pulled from [this transaction on the Sovrin
+MainNet](https://indyscan.io/tx/SOVRIN_MAINNET/domain/140326).
+
+``` json
+"data": {
+    "revocDefType": "CL_ACCUM",
+    "revocRegDefId": "4xE68b6S5VRFrKMMG1U95M:4:4xE68b6S5VRFrKMMG1U95M:3:CL:59232:default:CL_ACCUM:4ae1cc6c-f6bd-486c-8057-88f2ce74e960",
+    "value": {
+        "accum": "21 116...567",
+        "prevAccum": "21 128...C3B",
+        "issued": [
+        ],
+        "revoked": [
+            172
+        ]
+    }
+},
+```
+
+In the above:
+
+- `revocDefType`: is defined by the comparable entry in the [[ref: RevReg]], and for this
+  version of AnonCreds is hard-coded to `CL_ACCUM`.
+- `revocRegDefId`: is the Id of the [[ref: RevReg]] to which this entry is
+  being added.
+- `accum`: is the new value of the accumulator based on the state of the
+  credentials in the [[ref: RevReg]], including those listed in this transaction. The
+  value is calculated by the Issuer based on the credential state changes, and
+  submitted as part of the transaction, and verified by the ledger before being
+  published.
+- `prevAccum`: is the previous value of the accumulator. The value is supplied
+  by the Issuer in the transaction request, and verified by the ledger before
+  being published.
+- `issued`: an array (possibly empty or not supplied) of the indices within the
+  [[ref: RevReg]] of the credentials whose state has changed to `active` (also known as
+  `not revoked`) since the last [[ref: RevRegEntry]] was published to the ledger.
+- `revoked`: an array (possibly empty or not supplied) of the indices within the
+  [[ref: RevReg]] of the credentials whose state has changed to `revoked` (also known as
+  `not active`) since the last [[ref: RevRegEntry]] was published to the ledger.
+
+In the example transaction above no credentials are `issued` (meaning changed
+from status `revoked` to `issued`) and only one, the credential with index
+`172`, is changed to `revoked`. Both lists can have an arbitrary number of
+entries, up to the total number of credentials in the [[ref: RevReg]].
+
+The algorithm to calculate the value of a [[ref: RevRegEntry]] accumulator at
+any time is the same: determine the (modulo) product of the primes for each
+non-revoked credential in the [[ref: REV_REG]], as described
+[here](data_flow_setup.md#creating-the-initial-revocation-registry-entry-object).
+
+A [[ref: VDR]] publishing the [[ref: RevRegEntry]] transactions MAY perform its own calculation
+of the accumulator to ensure that the calculation of the accumulator after all
+of the revocation status updates to the credentials within the [[ref: RevReg]] have been
+applied, rejecting the transaction if the calculated accumulator does not match that from the [[ref: issuer]].
+
+The [[ref: issuer]] MUST track of the revocation status of all of the credentials
+within a [[ref: RevReg]] so that it can both calculate the correct accumulator and send
+to the [[ref: VDR]] accurate lists (`issued` and `revoked`) of the indices of the
+credentials whose status has changed since the last [[ref: RevRegEntry]] was published.
+If the [[ref: issuer]] and [[ref: VDR]] get out of sync about the status of
+credentials in the [[ref: RevReg]], such that the two cannot agree on the current value
+of the accumulator, the [[ref:issuer]] must rationalize the differences and
+produce a [[ref: RevRegEntry]] transaction that accounts for both the last published
+[[ref: RevRegEntry]] published in the [[ref: VDR]] and the desired revocation status of all of the
+credentials in the [[ref: RevReg]].
+
+Note that the [[ref: holder]] is not involved in the credential revocation
+process. There is no technical requirement for an [[ref: issuer]] to notify the
+[[ref: holder]] that a credential they were issued has been revoked. That said, it is a
+courtesy that may improve the user experience of the [[ref: holder]]. [Aries RFC 0183
+Revocation
+Notification](https://github.com/hyperledger/aries-rfcs/tree/main/features/0183-revocation-notification)
+is an example of how that can be done. Even if not notified by the [[ref:
+issuer]] of the revocation of a credential, the [[ref: holder]] can detect their
+credential has been revoked when they retrieve the list of revoked credentials
+from the [[ref: VDR]] and discover the index of their credential in the list.
+
 #### AnonCreds Presentation with Revocation
+
+Creating an AnonCreds presentation is a two-step process, beginning with a
+request from the [[ref: verifier]] asking the [[ref: holder]] to include a proof of
+non-revocation (PoN-R) in the presentation, and then the [[ref: holder]] creating
+the PoN-R and including it in the presentation sent to the [[ref: verifier]]. Both steps
+are covered here.
+
+**NOTE**: Often in discussions about verifiable presentations, the term "[[ref: prover]]"
+is used to indicate the participant generating the request presentation.
+Throughout the Hyperledger Indy AnonCreds implementation the term `[[ref: prover]]` is
+used in the names of methods performed out by that participant. However, because
+in AnonCreds the [[ref: holder]] and the [[ref: prover]] are always the same entity, we'll use
+[[ref: holder]] to refer to the participant generating the requested presentation to
+emphasize that the same entity is both issued credentials and generating the
+presentations from those credentials.
+
+##### Verifier Revocation Interval Request
+
+The inclusion of a "Revocation Interval" by the [[ref: verifier]] in a presentation
+request is a minimal extension to the Presentation Request data format that is
+defined in the [Create Presentation Request](#create-presentation-request)
+section of this document. Notably the `non_revoked` data item is added:
+
+```json
+"non_revoked" : {
+    "from" : <epoch date>
+    "to" : <epoch date>
+}
+```
+
+The item may at the outer level of the presentation request, such that it
+applies to all attributes and predicates, or with any or all of the attributes/predicates,
+applying only to those specific attributes and/or predicates.
+
+The use of "interval" is intended to have the semantic of saying that the [[ref:
+verifier]] will accept a PoN-R from any point in the `from` to `to` interval.
+The intention is that by being as flexible as the business rules allow means
+that the [[ref: holder]] and/or the [[ref: verifier]] may have cached [[ref:
+VDR]] data such that they don't have to go to the [[ref: VDR]] to get additional
+[[ref: RevRegEntry]] data. In practice, the use of an interval here is not well
+understood and tends to cause confusion amongst presentation request developers.
+The AnonCreds community recommends always using matching `to` and `from` values
+(see [Aries RFC 0441 Present Proof Best
+Practices](https://github.com/hyperledger/aries-rfcs/tree/main/concepts/0441-present-proof-best-practices#semantics-of-non-revocation-interval-endpoints)).
+
+While one would expect the `from` value to be the current time ("Prove the
+credential is not revoked right now"), its use allows the [[ref: verifier]] to ask
+for a PoN-R sometime in the past. This addresses use cases such as "Prove that
+your car insurance policy was not revoked on June 12, 2021 when the accident
+occurred."
+
+##### [[ref: holder]] Generation of Proofs of Non-Revocation
+
+A [[ref: holder]] preparing an AnonCreds presentation must determine what, if any, proofs
+of non-revovcation (PoN-Rs) must be added to the presentation based on a
+combination of what is in the proof request, and what verifiable credentials are
+to be used in the presentation. As noted in the [previous
+section](#verifier-revocation-interval-request), the presentation request may have the
+`non-revoked` item at the outer-most level, applying to all
+source credentials, or at the `requested_attribute` and/or
+`requested_predicate` level, applying only to specific source credentials. For each,
+the [[ref: holder]] must also determine if the verifiable credential selected for
+attributes/predicates where a PoN-R is requested is a revocable credential.
+Obviously, a PoN-R cannot be produced for a verifiable credential issued without
+a [[ref: RevReg]].
+
+Once the [[ref: holder]] has determined the required PoN-Rs needed for the presentation,
+they must generate a PoN-R for each applicable source verifiable credential and
+add the PoN-Rs to the presentation. For each, the [[ref: holder]] must collect the
+applicable [[ref: RevReg]] data published by the [[ref: issuer]] and then generate the proof.
+
+###### Collecting [[ref: RevRegEntry]] Data
+
+In order to produce a PoN-R, the [[ref: holder]] must collect the following information from wherever the [[ref: issuer]]
+has published the information. Note that the [[ref: holder]] may have some or all of this information cached
+from data previously collected.
+
+- The type of `issuance_type` of the [[ref: RevReg]] -- whether the initial state of the
+  credentials in the registry is `active` or `revoked`. This information is part
+  of the [[ref: RevReg]], and so likely cached by the [[ref: holder]] when they were first issued
+  the credential.
+- The tails file for [[ref: RevReg]], the location (a URL) of which is stored in the
+  issued credential's [[ref: RevReg]]. The [[ref: holder]] likely (though not necessarily) would
+  have collected the tails file at the time of issuance. Recall (from this
+  section of the specification) that the tails file for a [[ref: RevReg]] is generated at
+  creation time and does not change.
+- The index of the credential within the [[ref: RevReg]] for the [[ref: holder]]'s specific
+  credential being used in the presentation. This information is given to the
+  [[ref: holder]] by the [[ref: issuer]] when the verifiable credential is issued.
+- The accumulator published by the [[ref: issuer]] for the [[ref: RevRegEntry]] that the [[ref: holder]]
+  will use in generating the PoN-R. In the Hyperledger Indy implementation of
+  AnonCreds, the entries are published on the ledger, and collected via a
+  special request to the ledger (detailed below).
+- The revocation status changes of all of the credential indices up to the
+  publication of the accumulator that the [[ref: holder]] will use in generating the
+  proof. Required is the collection of all of the `issued` and `revoked` lists
+  (as described [here](#anoncreds-credential-revocation-and-publication)) from
+  all of the [[ref: RevRegEntry]] publication requests made by the [[ref: issuer]] up to and
+  including the request that includes the accumulator being used by the [[ref: holder]]
+  in generating the PoN-R.
+
+The collection of the last two items is difficult without extra support of the
+entity holding the published [[ref: RevReg]] (e.g. the [[ref: VDR]]/ledger).
+Since each [[ref: RevRegEntry]] holds only the list of `active` and `revoked`
+credential revocation status changes since the previous [[ref: RevRegEntry]]
+(the "deltas"), a [[ref: holder]] must retrieve those lists from every [[ref:
+RevRegEntry]] from [[ref: RevReg]] creation to the [[ref: RevRegEntry]] holding
+the accumulator the [[ref: holder]] will use for generating the PoN-R. The
+[[ref: issuer]] could have made many calls to publish [[ref: RevRegEntry]]
+transactions, and the [[ref: holder]] would have to make a request for each one,
+which is not practical (and perhaps not even possible). In the Hyperledger Indy
+implementation, a special call
+([`get_revoc_reg_delta`](https://github.com/hyperledger/indy-node/blob/master/docs/source/requests.md#get_revoc_reg_delta))
+is used to collect the necessary data from all the [[ref: RevRegEntry]]
+transactions for a specified interval in a single request. In the most used
+version of the call, the interval in the request is `from` 0 (meaning from when
+the [[ref: RevReg]] was created) `to` the current time. If the [[ref: holder]]
+has called the routine previously with an earlier `to` value and cached the
+results, the [[ref: holder]] MAY use the time of the cached result as the
+`from`, so that only the credentials with revocation status changes since that
+time are returned. The [[ref: holder]] then adds the returned lists to the
+cached lists. If the [[ref: verifier]] has requested a "back in time" PoN-R, the
+[[ref: holder]] may use a `to` date to match the date of interest to the [[ref:
+verifier]]. When executed, the transaction returns:
+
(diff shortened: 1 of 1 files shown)


swcurran:
Does not include the cryptographic steps to generate or verify the proof, nor the data in the "non_revoc_proof" section of a presentation. All of that needs to be added, once this is merged.

spec/data_flow_revocation.md
```diff
@@ -1,11 +1,429 @@
...
+section of this document. Notably the `non_revoked` data item is added:
+
+```json
+"non_revoked" : {
+    "from" : <Unix Epoch time>
+    "to" : <Unix Epoch time>
+}
+```
+
+The item may at the outer level of the presentation request, such that it
```
wip-abramson, line 175:
```suggestion
The item may be at the outer level of the presentation request, such that it
```

wip-abramson: sense check paragraph

spec/data_flow_revocation.md
```diff
@@ -1,11 +1,429 @@
...
+  section of the specification) that the tails file for a [[ref: RevReg]] is generated at
+  creation time and does not change.
+- The index of the credential within the [[ref: RevReg]] for the [[ref: holder]]'s specific
+  credential being used in the presentation. This information is given to the
+  [[ref: holder]] by the [[ref: issuer]] when the verifiable credential is issued.
+- The accumulator published by the [[ref: issuer]] for the [[ref: RevRegEntry]] that the [[ref: holder]]
+  will use in generating the NRP. In the Hyperledger Indy implementation of
+  AnonCreds, the entries are published on the ledger, and collected via a
+  special request to the ledger (detailed below).
+- The revocation status changes of all of the credential indices up to the
```
wip-abramson, line 237: delete `of`? Not sure. Sense check para

spec/data_flow_revocation.md
```diff
@@ -1,11 +1,429 @@
...
+presentation, including any PoNR(s), fail cryptographic verification. The
+verification code MAY surface additional detail about what part of the
+presentation failed, such as which NRP failed verification (if any).
+
+The [[ref: verifier]] SHOULD evaluate the presentation to make sure that the
+[[ref: holder]] provided all requested NRPs. Notably, if any expected PoNRs
+are not received in the presentation, the [[ref: verifier]] SHOULD check to see
+if the given credential type is revocable. If not, it is acceptable that no
+NRP was received. However, if the credential used in the generation of the
+proof is revocable, and the [[ref: holder]] did not provide the P0N-R, the
```
wip-abramson, line 427:
```suggestion
proof is revocable, and the [[ref: holder]] did not provide the NRP, the
```

wip-abramson:
Reads well to me! Just made a few minor suggestions

spec/data_flow_revocation.md
```diff
@@ -1,11 +1,429 @@
...
 #### AnonCreds Issuance with Revocation
 
+The details of an [[ref: issuer]] issuing a revokable credential to a [[ref:
+holder]] are covered in the [issuance data
+flow](data_flow_issuance.md#issue-credential) section of this specification.
+
+If the [[ref: issuer]] has created the [[ref: RevReg]] from which they are
+issuing credentials to have an `ISSUANCE_TYPE` of `ISSUE_ON_DEMAND`, the
+[[ref:issuer]] must publish a [[ref: RevRegEntry]] (as described in the next
+section) as each credential issued. Based on the experience of the AnonCreds
```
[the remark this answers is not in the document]
swcurran: Instead changed to "as each credential is issued."

spec/data_flow_revocation.md
swcurran: Reviewed and clarified, with missing word added.

spec/data_flow_revocation.md
```diff
@@ -1,11 +1,429 @@
...
+    {
+        "schema_id": "7BPMqYgYLQni258J8JPS8K:2:degree schema:46.58.87",
+        "cred_def_id": "7BPMqYgYLQni258J8JPS8K:3:CL:70:faber.agent.degree_schema",
+        "rev_reg_id": "7BPMqYgYLQni258J8JPS8K:4:7BPMqYgYLQni258J8JPS8K:3:CL:70:faber.agent.degree_schema:CL_ACCUM:61d5a381-30be-4120-9307-b150b49c203c",
+        "timestamp": 1656269796
+    }
+]
+```
+
+As always, the [[ref: holder]]'s presentation, with the embedded NRP(s), is set to the
```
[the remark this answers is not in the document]
swcurran: fixed.

spec/data_flow_revocation.md
```diff
@@ -1,11 +1,429 @@
...
+the presentation for the NRP being processed, plus the accumulator from
+appropriate [[ref: RevRegEntry]], the following steps are carried out to verify
+the NRP.
+
+:::todo
+To Do: Outline the NRP verification calculation.
+:::
+
+:::todo
+To Do: Is there a separate process to bind the NRP to the credential?
```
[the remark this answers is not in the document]
swcurran: Not sure -- I think this needs to be an issue that is investigated. I'll add it.

spec/data_flow_revocation.md
```diff
@@ -1,11 +1,429 @@
...
+:::
+
+The verification code MUST surface to the [[ref: verifier]] if any part of the
+presentation, including any PoNR(s), fail cryptographic verification. The
```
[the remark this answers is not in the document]
swcurran: fixed.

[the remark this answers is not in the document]
swcurran: FYI - I'm making the fix in my copy vs. accepting the commit because I'm not sure how sign off is done. I don't want to screw up the DCO.  Someday I'll try it... :-)

spec/data_flow_revocation.md
swcurran: Fixed.

swcurran:
Reworking section to move present proof components to the presentation section of the specification.

swcurran:
Updated per discussion at meeting on 2022.07.11. Migrated the sections related to issuing, presentation and verification into those parts of the specification. Retained the overall section to provide a summary of all parts of revocation in one place (with pointers to where the details are in other parts of the specification) and to cover the action of revoking (or activating) credentials.

Underlined the recommendation against using the "ISSUE_ON_DEMAND" option for issuing, where you have to update the RevReg whenever you issue a credential. Added a correlation danger for cases where "ISSUE_ON_DEMAND" is used and the credential has an "issue_date" type field in the credential. Verifiers knowing the issue_date can look on the ledger for a credential activated at that time and thus learn an identifier for the credential and monitor the revocation status of the credential.

@ArPhil -- please review -- particularly about the distribution of the revocation content into the other sections of the spec.

I'd like to get this merged so I can do some wholesale moving around of header levels in the specification -- move each of the issue, present, verify and revocation sections up one level.

spec/data_flow_revocation.md
```diff
@@ -1,11 +1,173 @@
...
-:::
+AnonCreds includes a mechanism that supports the revocation of verifiable
+credentials. This mechanism includes:
+
+- An [[ref: issuer]] setting up to issue revocable credentials.
+- An [[ref: issuer]] issuing revocable credentials.
+- An [[ref: issuer]] activating or revoking issued credentials.
+- A [[ref: verifier]] requesting a presentation to include a non-revocation proof
+  for one or more revocable credentials.
+- A [[ref: holder]] generating based on the request of the verifier a
```
ArPhil, line 11:
```suggestion
- A [[ref: holder]] generating based on the presentation request of the verifier a
```

spec/data_flow_revocation.md
```diff
@@ -1,11 +1,173 @@
...
+  non-revocation proof for attributes derived from revocable credentials.
+- A [[ref: verifier]] verifying a non-revocation proof included in a
+  presentation from a [[ref: holder]].
+
+A fundamental goal of AnonCreds is to not provide a correlatable identifier for
+a either a [[ref: holder]] or a credential as part of generation and verification of an AnonCreds
```
ArPhil, line 17:
```suggestion
either a [[ref: holder]] or a credential as part of generation and verification of an AnonCreds
```

spec/data_flow_revocation.md
```diff
@@ -1,11 +1,173 @@
...
+presentation. Applying that goal to revocation means that the revocation
+mechanism must support the [[ref: holder]] proving a credential used in generating a
+presentation is not revoked without providing a correlatable identifier for that
+credential. As such, the AnonCreds revocation mechanism uses a Zero Knowledge
```
ArPhil, line 21:
```suggestion
credential or the holder itself. As such, the AnonCreds revocation mechanism uses a Zero Knowledge
```

spec/data_flow_revocation.md
```diff
@@ -1,11 +1,173 @@
...
+Proof (ZKP) that allows the [[ref: holder]] to prove a credential they hold is not
+revoked without revealing an identifier for their credential.
```
ArPhil, line 23:
```suggestion
revoked without revealing an identifier for their credential or the holder.
```

spec/data_flow_revocation.md
```diff
@@ -1,11 +1,173 @@
...
+Further, if a credential contains some kind of "Issue Date" attribute in the
+credential, verifiers could use that attribute information combined with the
+publication of a [[ref: RevRegEntry]] at the same time to learn the ID of the
+holder's credential, giving them both a correlatable identifier for the holder
+and a way to monitor if the holder's credential is revoked.
+
+For these reasons we anticipate the deprecation or removal of the
+`ISSUE_ON_DEMAND` approach in the next version of AnonCreds specification.
+Feedback from the community on this would be appreciated. We are particularly
+interested in understanding what use cases there are for `ISSUE_ON_DEMAND`.
```
ArPhil, line 59:
```suggestion
::: warning
If the [[ref: issuer]] has created the [[ref: RevReg]] from which they are
issuing credentials to have an `ISSUANCE_TYPE` of `ISSUE_ON_DEMAND`, the
[[ref:issuer]] must publish a [[ref: RevRegEntry]] (as described in the [next
section](#anoncreds-credential-activationrevocation-and-publication)) as each
credential is issued. Based on the experience of the AnonCreds community in the
use of revocable credentials, it is highly recommended the `ISSUE_ON_DEMAND`
approach **NOT** be used unless absolutely required by your use case. The reason
for this recommendation is that `ISSUE_ON_DEMAND` requires near real-time
activation, meaning the issuing of each credential requires the
near-simultaneous publication of a new [[ref: RevRegEntry]], resulting in many
[[ref: RevRegEntry]] transactions being performed, one per credential issued.

Further, if a credential contains some kind of "Issue Date" attribute in the
credential, verifiers could use that attribute information combined with the
publication of a [[ref: RevRegEntry]] at the same time to learn the ID of the
holder's credential, giving them both a correlatable identifier for the holder
and a way to monitor if the holder's credential is revoked.

For these reasons we anticipate the deprecation or removal of the
`ISSUE_ON_DEMAND` approach in the next version of AnonCreds specification.
Feedback from the community on this would be appreciated. We are particularly
interested in understanding what use cases there are for `ISSUE_ON_DEMAND`.
:::
```

ArPhil: Added a "warning" highlight box, since we do not recommend the usage of `ISSUE_ON_DEMAND`. I am also wondering, if we should not move this entire paragraph to the credential definition setup part with revocation capabilities. 

We than have to distinguish the optional revocation part within each section into a) `ISSUANCE_PER_DEFAULT` and b) `ISSUE_ON_DEMAND`. 

Or we only mention the two issuing modes once at the credential definition setup part and emphasize, that *ISSUE_ON_DEMAND' probably will not be supported in future versions and therefore don`t explain the details. 

Not sure yet what the best approach here is...

spec/data_flow_revocation.md
```diff
@@ -1,11 +1,173 @@
...
+RevRegEntry]]](data_flow_setup.md#creating-the-initial-revocation-registry-entry-object)
+for a [[ref: RevReg]]. In that process, the accumulator for the initial state of
+the [[ref: RevReg]] is published. When subsequent [[ref: RevRegEntry]]
+transactions are published to the ledger, each includes an updated value of the
+accumulator, along with lists of the identifiers of the credentials that have
+had their revocation state toggled -- either changed to issued, or changed to
+revoked -- since the last [[ref: RevRegEntry]] was published. An example of the
+data in the [[ref: RevRegEntry]] is shown in the following example of a [[ref:
+RevRegEntry]], pulled from [this transaction on the Sovrin
+MainNet](https://indyscan.io/tx/SOVRIN_MAINNET/domain/140326).
```
ArPhil, line 76:
```suggestion
accumulator. The update of the accumulator is necessary with each revocation or activation of a credential or set of credentials since the last published [[ref: RevRegEntry]]. This is because only the factors (all factors are listed in the respective tails file) of credentials which are active (meaning not being revoked) contribute to the accumulator. Therefore in addition to the updated accumulator value, every RevRegEntry contains a list of indexes of credential factors which have been revoked within the [[ref: RevRegEntry]]. This list of factor indexes is a so called [[ref: Witness Delta]] and enables the [[ref: Holder]] to successfully generate a proof of non revocation.

An example of the
data in the [[ref: RevRegEntry]] is shown in the following example of a [[ref:
RevRegEntry]], pulled from [this transaction on the Sovrin
MainNet](https://indyscan.io/tx/SOVRIN_MAINNET/domain/140326).
```

ArPhil: I tried to differentiate between "identifiers of the credentials" and the "index of the factor within a tails file". From my point of view, this is crucial, since imho it is not some kind of ID of a credential, that is published within a RevRegEntry, but rather the index of the factor, that is bound to the credential. 

But I am not entirely happy with my wording here...

spec/data_flow_revocation.md
```diff
@@ -1,11 +1,173 @@
...
+
+The algorithm to calculate the value of a [[ref: RevRegEntry]] accumulator at
+any time is the same: determine the (modulo) product of the primes for each
+non-revoked credential in the [[ref: REV_REG]], as described
+[here](#creating-the-initial-revocation-registry-entry-object).
+
+A [[ref: VDR]] publishing the [[ref: RevRegEntry]] transactions MAY perform its own calculation
+of the accumulator to ensure that the calculation of the accumulator after all
+of the revocation status updates to the credentials within the [[ref: RevReg]] have been
+applied, rejecting the transaction if the calculated accumulator does not match that from the [[ref: issuer]].
```
ArPhil, line 128: Sounds wise to me, that the VDR does a double check on the updated accumulator value, but do we want to mention it here (in the AnonCreds Spec) or is it out of scope?

spec/data_flow_revocation.md
```diff
@@ -1,11 +1,173 @@
...
+Note that the [[ref: holder]] is not involved in the credential revocation
+process. There is no technical requirement for an [[ref: issuer]] to notify the
+[[ref: holder]] that a credential they were issued has been revoked. That said, it is a
+courtesy that may improve the user experience of the [[ref: holder]]. [Aries RFC 0183
+Revocation
+Notification](https://github.com/hyperledger/aries-rfcs/tree/main/features/0183-revocation-notification)
+is an example of how that can be done. Even if not notified by the [[ref:
+issuer]] of the revocation of a credential, the [[ref: holder]] can detect their
+credential has been revoked when they retrieve the list of revoked credentials
+from the [[ref: VDR]] and discover the index of their credential in the list.
```
ArPhil, line 150: Sounds wise to me, that the VDR does a double check on the updated accumulator value, but do we want to mention it here (in the AnonCreds Spec) or is it out of scope?

I think the same about getting the VDR and issuer revocation list syncing and about informing the user about that his credential has been revoked. On the las one we are rather on the Aries level, than the Indy / AnonCreds level imho?

spec/data_flow_revocation.md
```diff
@@ -1,11 +1,173 @@
...
+
+#### AnonCreds Verification with Revocation
+
+A [[ref: verifier]] receives the presentation from the [[ref: holder]] and
+processes the [non-revocation-related parts of the presentation](#verify-presentation) and
+the [revocation-related parts of the presentation](#verify-non-revocation-proof)
+(if any) in the presentation. The resulting status of the presentation combines the
+verification outcomes from processing all proofs within the presentation. If 
+verification of one or more of the embedded proofs is unsuccessful, the
+presentation is rejected as unverifiable.
```
ArPhil, line 173: Personally, I would move these sections to the "Create Presentation Request" and "Verify Presentation" sections.

spec/data_flow_presentation_create_presentation.md
```diff
@@ -1,15 +1,31 @@
...
+
+**NOTE**: Often in discussions about verifiable presentations, the term "[[ref: prover]]"
+is used to indicate the participant generating the request presentation.
+Throughout the Hyperledger Indy AnonCreds implementation the term `[[ref: prover]]` is
+used in the names of methods performed out by that participant. However, because
+in AnonCreds the [[ref: holder]] and the [[ref: prover]] are always the same entity, we'll use
+[[ref: holder]] to refer to the participant generating the requested presentation to
+emphasize that the same entity is both issued credentials and generating the
+presentations from those credentials.
+
```
ArPhil, line 27:
```suggestion
::: note
Often in discussions about verifiable presentations, the term "[[ref: prover]]"
is used to indicate the participant generating the request presentation.
Throughout the Hyperledger Indy AnonCreds implementation the term `[[ref: prover]]` is
used in the names of methods performed out by that participant. However, because
in AnonCreds the [[ref: holder]] and the [[ref: prover]] are always the same entity, we'll use
[[ref: holder]] to refer to the participant generating the requested presentation to
emphasize that the same entity is both issued credentials and generating the
presentations from those credentials.
:::
```

spec/data_flow_presentation_create_presentation.md
```diff
@@ -353,8 +369,193 @@ The resulting presentation `proof_json` created by the Holder has the following
...
+credentials, or at the `requested_attribute` and/or `requested_predicate` level,
+applying only to specific source credentials. For each, the [[ref: holder]] must
+also determine if the verifiable credential selected for attributes/predicates
+where a NRP is requested is a revocable credential. Obviously, a NRP cannot be
+produced for a verifiable credential issued without a [[ref: RevReg]].
+
+Once the [[ref: holder]] has determined the required NRPs needed for the presentation,
+they must generate a NRP for each applicable source verifiable credential and
+add the NRPs to the presentation. For each, the [[ref: holder]] must collect the
+applicable [[ref: RevReg]] data published by the [[ref: issuer]] and then generate the proof.
```
ArPhil, line 390:
```suggestion
applicable [[ref: RevReg]] respectively [[ref: RevRegEntry]]s data published by the [[ref: issuer]] and then generate the proof.
```

spec/data_flow_presentation_create_presentation.md
```diff
@@ -1,15 +1,31 @@
...
+multiple issuers, which should reside in the Holder's wallet.
 
-##### Protocol description
+The [[ref: verifier]] may specify in the presentation request that some or
+all of the attributes/predicates that are derived from revocable verifiable credentials held by
+the [[ref: holder]] have an accompanying non-revocation proof (NRP). The generation of an NRP
+is described [in this section](#generate-non-revocation-proofs) of the specification.
+
+**NOTE**: Often in discussions about verifiable presentations, the term "[[ref: prover]]"
+is used to indicate the participant generating the request presentation.
```
ArPhil, line 20:
```suggestion
is used to indicate the participant generating the presentation request response.
```

spec/data_flow_presentation_create_presentation.md
```diff
@@ -353,8 +369,193 @@ The resulting presentation `proof_json` created by the Holder has the following
...
+
+- The type of `issuance_type` of the [[ref: RevReg]] -- whether the initial state of the
+  credentials in the registry is `active` or `revoked`. This information is part
+  of the [[ref: RevReg]], and so likely cached by the [[ref: holder]] when they were first issued
+  the credential.
+- The tails file for [[ref: RevReg]], the location (a URL) of which is stored in the
+  issued credential's [[ref: RevReg]]. The [[ref: holder]] likely (though not necessarily) would
+  have collected the tails file at the time of issuance. Recall (from this
+  section of the specification) that the tails file for a [[ref: RevReg]] is generated at
+  creation time and does not change.
```
ArPhil, line 406:
```suggestion
  creation time and does never change.
```

spec/data_flow_presentation_create_presentation.md
```diff
@@ -353,8 +369,193 @@ The resulting presentation `proof_json` created by the Holder has the following
...
+- The index of the credential within the [[ref: RevReg]] for the [[ref: holder]]'s specific
+  credential being used in the presentation. This information is given to the
+  [[ref: holder]] by the [[ref: issuer]] when the verifiable credential is issued.
```
ArPhil, line 409: Since the heading of this section says "Collecting RevRegEntry Data", I am not sure if the holders credential index should be listed here as well, because it also says that the section is about information which was published by the issure earlier.

Maybe we have either to change the heading to be more generic (e.g. "Required information for NRP generation") or place this peace on information somewhere else in order to avoid confusion?

spec/data_flow_presentation_create_presentation.md
```diff
@@ -353,8 +369,193 @@ The resulting presentation `proof_json` created by the Holder has the following
...
+- The accumulator published by the [[ref: issuer]] for the [[ref: RevRegEntry]] that the [[ref: holder]]
+  will use in generating the NRP. In the Hyperledger Indy implementation of
+  AnonCreds, the entries are published on the ledger, and collected via a
```
ArPhil, line 412:
```suggestion
  AnonCreds, the entries are published as [[ref: RevRegEntry]]s on the ledger, and collected via a
```

spec/data_flow_presentation_create_presentation.md
```diff
@@ -353,8 +369,193 @@ The resulting presentation `proof_json` created by the Holder has the following
...
+
+- The full list of all `issued` and `revoked` entries in all of the [[ref: RevRegEntry]]
+  transactions within the requested interval.
+- The accumulator for the last [[ref: RevRegEntry]] within the requested interval.
+- The timestamp (in the Unix epoch format) of the last [[ref: RevRegEntry]] in the
+  interval.
+
+Once collected, the [[ref: holder]] processes the `issued` and `revoked` lists
+to determine the credential status (revoked or not) of every credential in the
+[[ref: RevReg]]. As well, the [[ref: holder]] can at the point see if the
```
ArPhil, line 454: Do we have an example oh how these lists look like exactly? Just wondering, whether the list entries have some timestamp which the holder has to use for "ordering" the entries and figuring out, whether a credential (which maybe as been activated, revoked and activated again) is now active or revoked

spec/data_flow_presentation_create_request.md
```diff
@@ -76,25 +76,58 @@ presents a decimal number (use `indy_generate_nonce` function to generate a corr
...
+
+The `restrictions` on each attribute/predicate defines a query filter in [[ref:
+Wallet Query Language]] (WQL) that is used to search the [[ref: holder]] wallet
+(storage) for candidate credentials that meet the restrictions. WQL is not
+required in an AnonCreds implementation, but the function it performs is
+required. That is, the [[ref: holder]] must scan the credentials it holds to
+find the ones that satisfy the requirements defined in the presentation request.
 
 The WQL specification can be found in
 [indy-sdk/docs/design/011-wallet-query-language/README.md](https://github.com/hyperledger/indy-sdk/blob/master/docs/design/011-wallet-query-language/README.md)
```
ArPhil, line 133: I somehow like, that the Wallet Query Language is mentioned and gives an idea how a holder (wallet) should be able to collect required credentials, but I am also not sure if we should mention it at all, because from point of view, that is outside the scope of AnonCreds. Just as you said: Some kind of mechanism is needed in order to be able to fetch required connections, but AnonCreds does not care (?) . :-)

spec/data_flow_presentation_verify.md
```diff
@@ -49,3 +54,52 @@ pub extern fn indy_verifier_verify_proof(command_handle: CommandHandle,
...
+To Do: Is there a separate process to bind the NRP to the credential?
+:::
+
+The verification code MUST surface to the [[ref: verifier]] if any part of the
+presentation, including any NRP(s), fail cryptographic verification. The
+verification code MAY surface additional detail about what part of the
+presentation failed, such as which NRP failed verification (if any).
+
+The [[ref: verifier]] SHOULD evaluate the presentation to make sure that the
+[[ref: holder]] provided all requested NRPs. Notably, if any expected PoNRs
```
ArPhil, line 98:
```suggestion
[[ref: holder]] provided all requested NRPs. Notably, if any expected NRPs
```
changed for consistency

swcurran:
Good stuff-- thanks!  I'll go through these and flag when this is ready for a (hopefully) final review.

spec/data_flow_revocation.md
swcurran: Fixed.

spec/data_flow_revocation.md
swcurran: Fixed.

spec/data_flow_revocation.md
swcurran: Fixed.

spec/data_flow_revocation.md
swcurran: I like the idea of mentioning it with the warning in the setup section and have moved it there.  I think I will leave a warning and pointer to the primary warning.

swcurran: Adjusted.  I moved the warning into the data setup, gave the warning a header, and then linked to that section in a couple of other places in the spec. I think that works pretty well.

spec/data_flow_revocation.md
swcurran: I consider the "RevocationID" to be the combined key `index+RevRegId`, and that the RevocationId is essentially a `CredentialId`.

**Question**: Does the `Witness Delta` include the factor contributed by the Credential that is the subject of the Non-Revocation Proof?  I've always wondered about that.  I think of it as the `Accumulator` = `Witness Delta` * `Credential Factor`

swcurran: I liked your wording and put it in verbatim. I think we can merge with that and adjust if needed.

spec/data_flow_revocation.md
swcurran: I think it is in scope as if the VDR (e.g. where ever it is published) and the issuer are not in sync, the holder will not be able to create a valid NRP. 

OK -- updated to reflect:

- If things get out of sync the NRP can't be created by the holder
- the VDR may prevent going out of sync
- the issuer must fix things if they do go out of sync.

I think that is reasonable to put into the spec.

spec/data_flow_revocation.md
swcurran: I moved the meat of the revocation sections over to the presentation and verification docs. I updated the wording to make it clear that this overview is only there to say -- "go look somewhere else".  I think that is reasonable.

spec/data_flow_presentation_create_presentation.md
swcurran: Thanks! I keep forgetting about that.

spec/data_flow_presentation_create_presentation.md
swcurran: Adjusted the wording.

spec/data_flow_presentation_create_presentation.md
swcurran: Fixed.

spec/data_flow_presentation_create_presentation.md
swcurran: Fixed and added link to tails section in the line above.

spec/data_flow_presentation_create_presentation.md
swcurran: I changed the title of the section -- good call!

spec/data_flow_presentation_create_presentation.md
swcurran: Fixed.

spec/data_flow_presentation_create_presentation.md
swcurran: The example in the data_flow_revocation.md file of a RevRegEntry includes the "lists".  Here is an example [RevRegEntry on Sovrin MainNet](https://indyscan.io/tx/SOVRIN_MAINNET/domain/140326).

spec/data_flow_presentation_create_presentation.md
```diff
@@ -353,8 +369,193 @@ The resulting presentation `proof_json` created by the Holder has the following
...
+process, perhaps with a notification to the [[ref: verifier]].
+
+###### Non-Revocation Proof Generation Steps
+
+Given the data collected by the [[ref: holder]] to produce the NRP, the
+following calculations are performed.
+
+A `witness` is calculated in the same way as the accumulator (as described
+[here](#publishing-the-initial-initial-revocation-registry-entry-object)),
+except the revocation status of the credential being proven as not revoked is
```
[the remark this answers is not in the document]
swcurran: Fixed.

spec/data_flow_presentation_create_request.md
swcurran: I tend to agree, but I didn't write that part. I think it can be removed in a subsequent PR.  Further, even if we do keep it, it should be moved to the generating presentation section, not in the create_request section.

But we'll leave it for now.

spec/data_flow_presentation_verify.md
```diff
@@ -49,3 +54,52 @@ pub extern fn indy_verifier_verify_proof(command_handle: CommandHandle,
 * `cb`: Callback that takes command result as parameter.
 * `Returns`
     * `valid`: true - if signature is valid, false - otherwise
+
+##### Verify Non-Revocation Proof
+
+If the presentation includes one or more Non-Revocation Proofs (NRPs) the
+[[ref: verifier]] must also extract from the verifiable presentation the NRPs
+and process each proofs. If any of the NRPs cannot be verified because one
```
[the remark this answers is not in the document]
swcurran: Fixed.

spec/data_flow_presentation_verify.md
swcurran: Fixed.

swcurran:
Great comments!  I've pushed an update with all of these addressed.  Some of the issues I've resolved, since the update was trivial, others I've left if you want to see what I have done.

spec/data_flow_revocation.md
ArPhil: > I consider the "RevocationID" to be the combined key `index+RevRegId`, and that the RevocationId is essentially a `CredentialId`.
> 
> **Question**: Does the `Witness Delta` include the factor contributed by the Credential that is the subject of the Non-Revocation Proof? I've always wondered about that. I think of it as the `Accumulator` = `Witness Delta` * `Credential Factor`

_This is handled by requiring accumulator updates to also publish a witness delta as part of the same transaction. This delta tells provers how to adjust their witness (referencing other indexes in the public tails file) to bring it back into harmony with the current value of the accumulator, such that the updated witness times the private factor once again equals the accumulator value._

(got it from [here](https://hyperledger-indy.readthedocs.io/projects/hipe/en/latest/text/0011-cred-revocation/README.html))

spec/data_flow_presentation_create_presentation.md
ArPhil: Thanks @swcurran !!

Looking at the `RevRegEntry` I see the transaction time besides the `revoked` and / or `issued` fields. But I also (somehwere read) that one can get all `RevRegEntry`s for a `RevReg` with only one request against the VDR. Not sure If I understood this correctly, but if so, the response has to contain either

a) an ordered list with all `RevRegEntries` (`revoked` and `issued` lists) so that the hilder / prover is able to figure out which credentials are / have been (re)activated and therefore knows which factors to consider for building the witness or

b) Only the (final) list of factors / indices of credential factors which contribute to the accumulator (the logic of figuring out, which factors have to be considered for building the current witness is done by the VDR in this case.

Maybe I have to look / read further for this...

spec/data_flow_revocation.md
ArPhil: Sounds reasonable to me. I think I would add a `::: note` wrapper around it, just to indicate, that is some kind of "additional" info the reader gets here. Or maybe even a `::: warning` if we consider it as very important. 

To me the fact that the issuer has to keep things in sync is crucial because that is in the heart of AnonCreds. The info, that the VDR can assist not so much, because that depends on where (which VDR) AnonCreds are implemented in / to. How the issuer can fix when being out of sync (in Indy) sounds to me like an "problems / fixing" section. But maybe that is just me...

(Just trying to keep the AnonCreds part as much reduced as possible while adding "additional" separately) 😄

ArPhil (approved):

ArPhil (approved):

ArPhil:
@swcurran , I approved the PR, but my voice does not count in the AnonCreds github project, since I am not a reviewer with write access...  😅

spec/data_flow_revocation.md
swcurran: OK, so the witness is the product of all of the unrevoked factors excluding the one for the credential held by the holder. But the witness delta is exactly as you used it in your note above.  I've added the Witness Delta to the terminology section.

In looking at that reference, I also documented how the NRP ZKP is generated.  I hope it is enough detail, although I've noted that more detail is needed on generating the commitments.

spec/data_flow_revocation.md
swcurran: Good comments.  I've put a note wrapper on that as you suggest.

spec/data_flow_presentation_create_presentation.md
swcurran: Check out the [GET_REVOC_REG_DELTA](https://hyperledger-indy.readthedocs.io/projects/node/en/latest/requests.html#get-revoc-reg-delta) transaction. 

It's not clear how the lists are handled if an entry is both activated and revoked in the delta period with ordering the updates, but presumably the code is smart enough to ensure that from the lists, the holder can determine the correct revocation state of every credential in the registry.

swcurran:
Thanks -- at the next meeting, I'll propose that @ArPhil  be added as an Editor.

@TelegramSam @victormartinez-work -- could one of you review this PR and approve it?

I've added a couple of more tweaks based on @ArPhil 's comments.

TelegramSam (approved):
Discussed WG 20220725.

PR #55, opened 2022-07-04
Crypto section for Issuer PK gen (non revocation)

ale-linux:
This commit adds a section to the crypto specs detailing the issuer key gen
algorithm. This is only the credential keypair, the revocation keypair will
follow.

Signed-off-by: Alessandro Sorniotti <aso@zurich.ibm.com>

PR #56, opened 2022-07-08
Add crypto section to generated spec

swcurran:

The crypto.md file merged previously was not being referenced in the "specs.json" file, so was not being generated into the spec.  This change puts it into the spec.

~~In doing that, I discovered that the generated text looks pretty bad (red text...) as a block, so I added a type on the text block.~~

~~@ale-linux -- this will probably mess up your PR #55 because of the addition of the text blocks.~~

### DIFF
--- a/spec/crypto.md
+++ b/spec/crypto.md
@@ -18,7 +18,7 @@ _TODO: clarify exact format and encoding of inputs and outputs_
 
 `ProveCLCommit` finds has its reference implementation [here](https://github.com/hyperledger/ursa/blob/ece6ce32a59df4e1f99fa38243c1236423066bc2/libursa/src/cl/prover.rs#L1313-L1394).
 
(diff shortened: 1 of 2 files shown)


swcurran:
@ale-linux and others:

Not sure what type of code-block is being used for the cryptography section. With no name on the fenced block (e.g. just three backticks), it looks terrible -- just red text.  I found a bunch of languages that resulted in a black box containing fixed-font text, which looks reasonable.  I picked `tex`, but that's not right, but it looks OK.

Any suggestions on what the type of the text within the blocks should be?

swcurran:
Changed this PR to just add the crypto.md file to the spec.  Still need to sort out the fenced blocks.

TelegramSam (approved):
Discussed WG 20220711.

Issue #52, opened 2022-06-27
Add section to specification to describe how DIDs are used and not used in AnonCreds


swcurran:
Per the [WG meeting on 20220613](https://github.com/AnonCreds-WG/anoncreds-spec/wiki/WG-2022-06-13), add a section to the introduction(-ish) section to cover who DIDs are and are not used in AnonCreds. Suggestion is to not define the process, but to help understanding by talking about how DIDs really don't play into AnonCreds cryptography, but may have a role in both publishing objects, and in messaging amongst the participants.

swcurran:
Resolved #58

PR #59, opened 2022-07-10
Add definition for DIDs and how they are used in AnonCreds

swcurran:
Addresses https://github.com/AnonCreds-WG/anoncreds-spec/issues/52 -- documenting how DIDs are used AnonCreds.

Added by putting in a definition [[def: DID]] and then including in the definition a description of how DIDs are used. Not sure if that is right -- could move that to another section.

If you think the part about how DIDs are used/not used should move -- let me know where you think they should go. 😄

### DIFF
--- a/spec/data_flow_setup.md
+++ b/spec/data_flow_setup.md
@@ -69,22 +69,24 @@ sequenceDiagram
 
 Those with a knowledge of DIDs might expect that in the flow above, the first
 step would be for the [[ref: issuer]] to publish a DID. However, in AnonCreds,
-DIDs are not used in the processing of credentials, but rather are used to
-identify the entity publishing the objects used in the processing of credentials
--- the [[def: SCEHMA]], [[def: CRED_DEF]], [[def: REV_REG_DEF]] and [[def:
-REV_REG_ENTRY]] objects. That is, there is an enforced relationship between the
-DID of the entity publishing the AnonCred objects, and the objects themselves.
-For example, in the Hyperledger Indy implementation of AnonCreds, for a
-credential issuer to publish a [[def: CRED_DEF]] on an instance of Indy it must
-have a DID on that instance, and it must use that DID to sign the transaction to
-write the [[def: CRED_DEF]].
+DIDs are not used in the processing of credentials, and notably, the public keys
+used in AnonCreds signatures come not from DIDs, but rather from [[def:
+CRED_DEF]] objects. DIDs may be used to identify the entity publishing the
+objects that are then used in the processing of credentials -- the [[def:
+SCEHMA]], [[def: CRED_DEF]], [[def: REV_REG_DEF]] and [[def: REV_REG_ENTRY]]
+objects. There is an enforced relationship between an identifier (such as a DID)
+for the entity publishing the AnonCred objects, and the objects themselves. For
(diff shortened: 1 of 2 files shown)


swcurran:
Updated this to put the bulk of the "AnonCreds doesn't really use DIDs" explanation into the Setup Data Flow section, right under the initial sequence diagram.  Adjusted the `def` section accordingly.

I think this is ready to go -- @TelegramSam @victormartinez-work @hkny

spec/data_flow_setup.md
```diff
@@ -69,6 +65,39 @@ sequenceDiagram
...
+Further, the Hyperledger Indy ledger enforces, and makes available for
+verification, the requirement that the writing of the AnonCreds object must be
+signed by the DID that is writing the object.
+
+If a DID-based messaging protocol, such as
+[DIDComm](https://identity.foundation/didcomm-messaging/spec/) is used between
+the AnonCreds participants (the [[ref: issuer]], [[ref: holder]] and [[ref:
+verifier]]) the DIDs used for messaging are not used at all in AnonCreds. Such
+DIDs are used only to facilitate secure messaging between the participants to
+enable the issuing of credentials and the presentation of proofs.
```
hkny, line 97: I am not sure about this statement. In our current configuration, it seems like it is possible to use one DID for communications and credential issuance. Here is why I think it might not be the case.

1) In Aca-Py I create a new connection invitation with Public DID. The result is the following:
<html><head></head><body><span><span>

  | Response bodyDownload{   "connection_id": null,   "invitation": {     "@type": "did:sov:BzCbsNYhMrjHiqZDTUASHg;spec/connections/1.0/invitation",     "@id": "86a80a25-7f92-4560-a25a-0ed4bc4d2e5f",     "imageUrl": "https://routing.lissi.io/api/Image/lissiLogo",     "did": "did:sov:Lc4JLxaNPtrWmGXV5BsJKz",     "label": "default_tenant"   },   "invitation_url": "?c_i=eyJAdHlwZSI6ICJkaWQ6c292OkJ6Q2JzTlloTXJqSGlxWkRUVUFTSGc7c3BlYy9jb25uZWN0aW9ucy8xLjAvaW52aXRhdGlvbiIsICJAaWQiOiAiODZhODBhMjUtN2Y5Mi00NTYwLWEyNWEtMGVkNGJjNGQyZTVmIiwgImltYWdlVXJsIjogImh0dHBzOi8vcm91dGluZy5saXNzaS5pby9hcGkvSW1hZ2UvbGlzc2lMb2dvIiwgImRpZCI6ICJkaWQ6c292OkxjNEpMeGFOUHRyV21HWFY1QnNKS3oiLCAibGFiZWwiOiAiZGVmYXVsdF90ZW5hbnQifQ==" }
-- | --
</span></span></body></html>

2) With the same DID (Lc4JLxaNPtrWmGXV5BsJKz) I created a Schema Definition:
image

hkny: To me it seems like the same DID can be used for creating DIDComm invitations AND credential issuance

swcurran: Updating the phrasing to try to address those concerns. Let me know what you think.

TelegramSam (approved):
I like this. Thanks.

hkny (approved):
I am also happy with it.

PR #64, opened 2022-07-31
AnonCreds Objects Methods section of spec

swcurran:


swcurran:
Updated this section to add all that was discussed at the Working Group Meeting of 2022.08.01. Love to have a review from the those attending to see whether this captures it all.  Note that I didn't create the Registry as a table, as I wanted to put more text than could be managed in a markdown table.

spec/anoncreds_objects_methods.md
```diff
@@ -0,0 +1,128 @@
...
+Specification](https://www.w3.org/TR/did-core/), [[ref: AnonCreds Object
+Methods]] specify registration and resolution mechanisms for AnonCreds objects.
+The methods are added by their maintainers to the [AnonCreds
+Objects Method Registry](#anoncreds-objects-method-registry).
+
+[[ref: AnonCreds Objects Methods]] specify how issuers register AnonCreds
+objects ([[ref: Schemas]], [[ref: CredDefs]], [[ref: Rev_Reg_Defs]] and [[ref:
+Rev_Reg_Entrys]]), including the specification of the identifier format for each
+type of object, to what [[ref: Verifiable Data Registry]] the objects are
+published, how issuers register the objects, and how anyone can resolve the
```
ArPhil, line 18: ... specify how issuers register AnonCreds objects ([[ref: Schemas]] ... mentioned a few lines above already.

spec/anoncreds_objects_methods.md
```diff
@@ -0,0 +1,128 @@
...
+identifiers to retrieve the published objects. Implementations of the AnonCreds
+specification should be organized so as to support for issuers at least one
+[[ref: AnonCreds Objects Methods]] for registration, and one or more [[ref:
+AnonCreds Objects Methods]] for resolution. AnonCreds issuers choose what
+AnonCred registration method(s) they use, and all AnonCreds participants choose
```
ArPhil, line 23: missing "s" in "AnonCred",... I guess?

spec/anoncreds_objects_methods.md
```diff
@@ -0,0 +1,68 @@
+## AnonCreds Objects Methods
+
+In the AnonCreds data flows are specifications of data models that contain
+identifiers to AnonCreds objects ([[ref: Schemas]], [[ref: CredDefs]], [[ref:
+Rev_Reg_Defs]] and [[ref: Rev_Reg_Entrys]]) that are published by issuers to
+locations ([[ref: Verifiable Data Registries]] or VDR) accessible to holders and
+verifiers. The format of the objects identifiers and the location of the objects
```
ArPhil, line 7: Should we mention, that these (public) objects are (and have to be) accessible by everybody and **in particular** by holders and verifiers, since they need these objects for issuance and presentation / verification?

spec/anoncreds_objects_methods.md
```diff
@@ -0,0 +1,68 @@
...
+Registry](#anoncreds-objects-method-registry).
+
+Each [[ref: AnonCreds objects method]] specifies the format of the object
+identifiers, to what [[ref: Verifiable Data Registry]] the objects are
+published, how issuers register (publish) objects, and how anyone can resolve
+the identifiers to retrieve the published objects. Implementations of an
+AnonCreds should be organized so as to allow issuers access to at least one
+[[ref: AnonCreds objects method]] for registration, and to allow holders and
+verifiers to access one or more [[ref: AnonCreds Objects Methods]] for
+resolution. AnonCreds issuers can then choose what AnonCreds registration
```
ArPhil, line 23: Since I missed the last meetings, maybe I also missed some discussion around this point, but I am not sure if I understand it correctly:

But when the issuer users "just" or "at least" one AnonCreds objects method, how can holders and verifiers be provided with access to one **or more** AnonCreds object methods for resolution? As far as I understand, the amount of AnonCreds objects methods usefull to the verifier and issuer only (?) depends on the amount of AnonCreds objects the issuer uses for issuance?

spec/anoncreds_objects_methods.md
```diff
@@ -0,0 +1,68 @@
...
+generated from [this
+repository](https://github.com/AnonCreds-WG/anoncreds-objects-methods-registry).
+
+The AnonCreds Objects Methods registry repository and published registry is
+managed by the AnonCreds Specification Working Group based on this [governance
+framework](https://anoncreds-wg.github.io/anoncreds-objects-methods-registry/#governance).
+
+Each entry in the [AnonCreds Objects Method
+Registry](https://anoncreds-wg.github.io/anoncreds-objects-methods-registry/)
+links to a specification for the associated [[ref: AnonCreds objects method]].
```
ArPhil, line 62: I had a look into the [AnonCreds object method template](https://github.com/AnonCreds-WG/anoncreds-objects-methods-registry/blob/main/registry/method_template.md) and as far as I understood, a AnonCreds object method entry should provide (only) some basic (contact) information including a link to the concrete specification of the AnonCreds object method itself.

I wonder:

- Should we also provide a template for a AnonCreds object method specification as well?
- In this template, should we list the "common" AnonCreds object method "functions" like "create schema", "create creddef", "update revocation_registry" and others so that an implementer can describe those "calls" to the VDR for the concrete implementation?
- Should we also provide a "security considerations" section in that template, since not all AnonCreds object methods provide the same guarantees?

I do not want to overcomplicate things, but I believe, that it would be more comfortable for interesting parties to have a common standard / schema for describing different implementations of AnonCreds object methods functions by using a similar / same document structure. This could improve the comparability of different AnonCreds object methods. Also it would provide a common abstract AnonCreds Object method API with all relevant functions an implementer has to implement. That is somehint that we do not have in the AnonCreds Specification yet, since we focused on the interaction flows but never listed the required calls against the VDR separately (in a clean list).

spec/anoncreds_objects_methods.md
swcurran: What I'm intending with this is that an issuer is likely to use one particular VDR for registering/publishing their objects, and the holders/verifiers will likely want to support one or more registries so that they can process verifiable credential from a variety of issuers.  The latter might even want to use a universal resolver, so as to (in theory) handle any AnonCreds Method.

swcurran: Cleaned up the language -- I think this is good to go.

spec/anoncreds_objects_methods.md
swcurran: Cleaned up.

swcurran:
I think this is ready to go, and getting this out there sooner than later would be helpful.

@ArPhil @TelegramSam @victormartinez-work -- could I get some approvals on this PR?

ArPhil:
@swcurran , your changes look good to me, but I still don`t have the proper permissions to review / approve the PR. Can update my role permissions?

victormartinez-work (approved):

ArPhil (approved):
Looks good to me. Thanks!

PR #65, opened 2022-08-01
Add some more context regarding VC/VP

c2bo:
- Attempt to add some more context about the general setup of credential/presentation
- Some additional terminology parts filled out

Still WiP, will likely change quite a bit

### DIFF
--- a/spec/terminology.md
+++ b/spec/terminology.md
@@ -116,7 +116,7 @@ accumulator, such that the updated witness times the private factor once again
 equals the accumulator value.
 
 [[def: Verifiable Data Registry]]
-~ TODO
+~ In order for a [[ref: DID]] to be resolvable, it is typically stored on some underlying system like a database or a network of some kind. Any such system that supports the basic functionality necessary for a [[ref: DID]] is called a Verifiable Data Registry (or in short: VDR) and the specification on the system including the steps necessary for operations on a DID is typically called a [DID-Method](https://www.w3.org/TR/did-core/#dfn-did-methods). Such a VDR can be realized based on different technologies with one example being a distributed ledger as is used in [did:indy](https://hyperledger.github.io/indy-did-method/).
 
 ::: todo
 Finalize glossary entries
--- a/spec/terminology.md
+++ b/spec/terminology.md
@@ -10,7 +10,17 @@
 ~ TODO
 
 [[def: credential]]
-~ A credential is a set of [[ref: claim]]s about an identity [[ref: subject]]. A verifiable credential is a tamper-proof credential whose authorship is cryptographically verifiable. An anonymous credential, also known as AnonCreds, is a verifiable credential that has privacy-preserving properties to enable data minimization and correlation resistance. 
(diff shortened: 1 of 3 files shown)


spec/terminology.md
```diff
@@ -10,7 +10,17 @@
...
+~ A credential is a set of [[ref: claim]]s about an identity [[ref: subject]]. A [[ref: Verifiable Credential]] is a tamper-proof credential whose authorship is cryptographically verifiable. An anonymous credential, also known as AnonCreds, is a verifiable credential that has privacy-preserving properties to enable data minimization and correlation resistance. 
+
+[[def: Verifiable Credential]]
+~ A verifiable credential is a temper-evident credential with cryptographically verifiable authorship. A verifiable credential is exchanged between a [[ref: issuer]] and a [[ref: holder]]. It consists of one or more [[ref: claims]] made by an [[ref: issuer]] about the [[ref: holder]] of a credential. One or several verifiable credentials can be used to construct a [[ref: Verifiable Presentation]] containing croptographically verifiable proofs of the [[ref: claims]]. In the case of the AnonCreds protocol, anonymous credentials known as AnonCreds are used that focus heavily on privacy features like [[ref: zero-knowledge proofs]].
```
wip-abramson, line 16:
```suggestion
~ A verifiable credential is a tamper-evident credential with cryptographically verifiable authorship. A verifiable credential is exchanged between a [[ref: issuer]] and a [[ref: holder]]. It consists of one or more [[ref: claims]] made by an [[ref: issuer]] about the [[ref: holder]] of a credential. One or several verifiable credentials can be used to construct a [[ref: Verifiable Presentation]] containing cryptographically verifiable proofs of the [[ref: claims]]. In the case of the AnonCreds protocol, anonymous credentials known as AnonCreds are used that focus heavily on privacy features like [[ref: zero-knowledge proofs]].
```

TelegramSam:
Discussed WG 20220801
- The word 'derived' is used in the VC spec.
- Multiple terms can be defined at the same time

spec/terminology.md
```diff
@@ -10,7 +10,17 @@
...
+
+~ A verifiable Credential depends on a [[ref: CRED_DEF]], a [[ref: SCHEMA]] and optionally a [[ref: REV_REG_DEF]] and corresponding [[ref: REV_REG_ENTRY]] (for credentials that support revocation) to be present on a [[ref: Verifiable Data Registry]] that is accessible to all participants of the ecosystem.
```
wip-abramson, line 18: Wondering if this should be AnonCred? A Verifiable Credential, is the term reserved by the W3C, we should be clear about the distinction.

AnonCreds are something that can be represented as VCs. But VCs do not depend on AnonCreds

wip-abramson:
Reviewing this it feels like we are trying to redefine common terms within the space, rather than just pointing to the common definitions. E.g. VDR, VC and VP.
 
Rather, I feel we should aim to explain the relation between AnonCreds and these existing definitions?

c2bo:
I agree with that assessment and have tried to rephrase the whole section to better explain the relation and differences between the 2. So far, I have not really come to a state that i am happy with.
I will close this PR and open a new one if I have a better version focusing on the AnonCreds aspects.

PR #66, opened 2022-08-01
reworked terminology section

ArPhil:
Signed-off-by: Artur Philipp <artur.philipp@posteo.de>


ArPhil:
Worked a bit on the terminology part:

- Added terminology for a) Credential Request b) Credential Offer c) Verifiable Data Registry
- Sorted terminology in lexigraphical order
- Remaned "Table of contets" to "Summary of contents", since we already have a ToC on the left pane in the doc. I`d rather like to have some summarizing text of the various sections of the spec here.
- Elaborated further on some parts like CRED_DEF.

swcurran (approved):
This looks good.  Note that:

- matching is case-insensitive, so it is OK to put the definition in a consistent format
- the definition can be a comma-separated list with synonyms following the main definition. The synonyms are NOT listed, so you can have `Claim, claims` to get the matches on `claims` as a reference.

I'm happy. to have this merged now, as we will do more passes on this. Or @ArPhil , you might want to adjust the way each entry is named and add some synonyms.

We don't want this PR open long as we have others that need it merged.

ArPhil:
@swcurran , thanks for the hint with the synonyms. I hope I catched all of the inconsistencies! :-)

Also added `Non-Revocation Proof (NRP)` to the terminology section. I believe that we can merge it now.

spec/terminology.md
```diff
@@ -1,25 +1,38 @@
...
 [[def: Correctness Proof]]
 ~ TODO
+::: todo
+Describe (key) Correctness Proof
+:::
+
+[[def: credential, credentials]]
+~ A credential is a set of [[ref: claims]] about an identity [[ref: subject]]. A verifiable credential is a tamper-proof credential whose authorship is cryptographically verifiable. An anonymous credential, also known as AnonCreds, is a verifiable credential that has privacy-preserving properties to enable data minimization and correlation resistance. 
+
+[[def: CRED_DEF, CDRED_DEFs]]
```
swcurran, line 15: Typo on this line -- CDRED.  Also, what about adding "CLAIM_DEF" as a synonym?

spec/terminology.md
```diff
@@ -1,25 +1,38 @@
...
 
-[[def: credential]]
-~ A credential is a set of [[ref: claim]]s about an identity [[ref: subject]]. A verifiable credential is a tamper-proof credential whose authorship is cryptographically verifiable. An anonymous credential, also known as AnonCreds, is a verifiable credential that has privacy-preserving properties to enable data minimization and correlation resistance. 
+~ A private_CRED_DEF is stored on the [[ref: issuer]]`s side and is an object that contains the private keys and the attributes related to a CRED_DEF. These private keys are used to create signature proofs for the issued anonymous credentials, which then can be validated in a derived form by a [[ref: verifier]]. A private_CRED_DEF never leaves the [[ref: Issuer]]'s domain and is stored securely.
+
+~ Revokable Verifiable Credentials require CRED_DEFs which also reference a [[ref: REV_REG_DEF]] (see below).
+
+:::todo
```
swcurran, line 25: Hmmm...maybe we make CRED_DEF_PRIVATE a symonym as well, and include it's definition here?  Or do we need a separate entry?

spec/terminology.md
```diff
@@ -30,86 +43,80 @@ publishers]] and [[ref: issuers]] and the AnonCreds objects they publish. This
...
+~ AnonCreds are bound to the [[ref: holder]] with a non-correlatable secret only known to the [[ref: holder]] itself called a link secret*. Instead of a persistent identifier, the link secret as a blind attribute is sent to the  [[ref: issuer]] during credential issuance. The issuer signs every calim (including the blinded link secret) individually, enabling [[ref: selective disclosure]] (see below). It means the [[ref: issuer ]] does not know the exact value of the link secret, and the [[ref: holder]] can prove the ownership of credentials to a [[ref: verifier]] without disclosing a persistent identifier.
```
swcurran, line 56: Ooppss...another typo -- calim.

swcurran (changes requested):
I found a couple of typos, so requested a change and added another random comment that you can change or ignore as you think best.

spec/terminology.md
ArPhil: Fixed Type. Added CLAIM_DEF as a synonym.

spec/terminology.md
ArPhil: Good point! At same sections in the spec I had the feeling, that the difference might not always be entirely clear to the reader. I added not all variants of synonyms to the CRED_DEF: 

`[[def: CRED_DEF, CRED_DEFs, CLAIM_DEF, CLAIM_DEFs, CRED_DEF_PRIVATE, CRED_DEF_PUBLIC]]`

I believe, that we should have one `CRED_DEF` in the terminology section, where we elaborete within that definition the difference between the public and private part of it. To me a `CRED_DEF` is one single thing, which is just divided into two parts, that are stored in different places (ledger and wallet). I would like to emphasize the cohesiveness by just providing one definition in the terminology section. 

But of course, we can also discuss this if you or others feel different about it.

spec/terminology.md
ArPhil: Sorry for that... but thanks for the good eye! :-)

spec/terminology.md
ArPhil: Could not resist and reworked the CRED_DEF part. Please tell me what you think about it.

swcurran (approved):
LGTM

swcurran:
Good stuff!

PR #68, opened 2022-08-01
Security Considerations

hkny:
Signed-off-by: Hakan Yildiz <c.hakan.yildiz@gmail.com>


spec/security_considerations.md
```diff
@@ -1,5 +1,36 @@
 ## Security Considerations
 
+### Cryptography
+
+#### Signature
+AnonCreds creation and validation 
+
+#### Revocation / Accumulators
+
+### Verifiable Data Registry
```
ArPhil, line 10: As discussed in the AnonCreds Specification meeting on 08.08.2022:

Would be nice to have some kind of AnonCreds Requirements checklist which can be used in order to verify, that the underlaying VDR is "compatible" / "suitable" for AnonCreds 1.0.

spec/security_considerations.md
```diff
@@ -1,5 +1,36 @@
...
+
 ::: todo
-Add security considerations.
+Do we want to add zeroMQ communication with Ledger / Nodes?
 :::
+
+#### Authentication of Issuer
+How does the verifier know that 
+
+### Envelope
```
hkny, line 24: Mention that there is a necessity to encrypt the credential in an encrypted way. Sufficiently encrypted but not necessarily mention a specific method

spec/security_considerations.md
```diff
@@ -1,5 +1,36 @@
...
+
+### Transport
```
hkny, line 26: same as envelope

spec/security_considerations.md
```diff
@@ -1,5 +1,36 @@
...
+
+### Wallet
+#### Recovery
+Seeds / DKMS etc.
+
+#### Support of Hardware Secure Modules
+
+#### Crypto Agility
+Post quantum crypto security
+
```
hkny, line 36: the signature algorithms might evolve but we must keep privacy-preserving features such as selective disclosure non-correlatability etc.

spec/security_considerations.md
```diff
@@ -1,5 +1,36 @@
 ## Security Considerations
 
+### Cryptography
+
+#### Signature
+AnonCreds creation and validation 
+
+#### Revocation / Accumulators
```
hkny, line 8: correlation in case of the revocation registries are too small.

spec/security_considerations.md
swcurran: Your phrasing triggered for me that this is needed in the AnonCreds Objects Method section -- what support is needed in a VDR in order for it to be used for AnonCreds 1.0?  Notably -- the ability to store AnonCreds objects, which are basically resources.

spec/security_considerations.md
```diff
@@ -1,5 +1,45 @@
...
+It is recommended to use a [[ref: Verifiable Data Registry]] that complies with the state of the art security features and best practices.
+
+#### Permission Management
+The VDR for storing [[ref: SCHEMA]]s, [[ref: CRED_DEF]]s, [[ref: REV_REG_DEF]]s, and [[ref: REV_REG_ENTRY]]s shall ensure that only the owner or from the owner permitted entities can write, edit, or revoke those AnonCreds objects. Furthermore, public AnonCreds objects shall be readable by any entity that can access the VDR.
+
+#### Authentication of Issuer
+The VDR shall allow only permitted entities to issue AnonCreds based on AnonCreds objects related to it.  
+
+#### Security Requirements at DIDs and Related Keys
+It is recommended to follow the best practices of decentralized key management system designs and architectures. The Hyperledger Indy implementation of AnonCreds are built on the following [DKMS Design and Architecture](https://github.com/hyperledger/indy-sdk/blob/main/docs/design/005-dkms/DKMS%20Design%20and%20Architecture%20V3.md)
```
swcurran, line 25: Suggest changing: "The Hyperledger Indy implementation of AnonCreds are built on..."

To: "For example, an option for publishing AnonCreds objects is Hyperledger Indy, which is built on..."

spec/security_considerations.md
```diff
@@ -1,5 +1,45 @@
...
+
+#### Authentication of Issuer
+The VDR shall allow only permitted entities to issue AnonCreds based on AnonCreds objects related to it.  
+
+#### Security Requirements at DIDs and Related Keys
+It is recommended to follow the best practices of decentralized key management system designs and architectures. The Hyperledger Indy implementation of AnonCreds are built on the following [DKMS Design and Architecture](https://github.com/hyperledger/indy-sdk/blob/main/docs/design/005-dkms/DKMS%20Design%20and%20Architecture%20V3.md) 
+
+### Envelope
+AnonCreds should be packed in a message envelope that can fulfill properties such as authenticity, integrity, confidentiality, and non-repudiation of the message. A message can have these properties with signature and encryption algorithms. It is recommended to choose signature and encryption algorithms that are state of the art and offer such security.
```
swcurran, line 28: Should we say that DIDComm is a good option for this?

swcurran (changes requested):
A couple of tweaks proposed, with the one about Hyperledger Indy being "an example" is the one that is the requested change.

spec/security_considerations.md
hkny: For example, Hyperledger Indy utilizes DIDComm v1 as message envelope for exchanging AnonCreds between [[ref: issuer]]s, [[ref: holder]]s, and [[ref: verifier]]s.

ArPhil:
I believe the current state is a great initial starting point for security considerations. I wonder, should we additionally consider the different perspectives of the roles (VDR, issuer, verifier, holder) on this topic? E.g. :

A holder requires:
- to protect the link secret
- to have options to authenticate other entities (especially the verifier) when sharing data
- ...

A Issuer requires:
- to protect the signing keys of credentials
- make sure, that the holder he wants to issue a credential to, is authentic
- ...

I checked the W3C DID Core and VC Data Model specs. Looks to me, that even there this separation of roles / concerns regarding security is not differentiated too much. But I believe it could help.

swcurran (approved):

PR #69, opened 2022-08-02
Remove Data Flow intro and Data Models sections and update headers levels in data flow sections

swcurran:


swcurran:
Per discussion at the Working Group call (2022-08-01) - removing the intro to the Data Flow section and bringing out one header level all of the Data Flow sections.  Per an earlier decision to **not** have a separate data model section, removed that still empty section of the spec.

Would like to have this merged quickly so that we can work on all the files changed here.

hkny (approved):

PR #71, opened 2022-08-05
Update the data flow sections to be ledger agnostic

swcurran:


swcurran:
The start of the work to make the data flow sections of the specification to be ledger agnostic.

swcurran:
I've gone through the rest of the data flow sections and I don't see any updates needed to make the spec "VDR-agnostic".  While there are references to the Object identifiers (e.g. schemaID and credDefID), none reference the format -- just that the IDs are needed.  That's appropriate. It is only during creation/publishing are the IDs created.

This is now good to go for reviewers -- @TelegramSam @ArPhil @hkny @victormartinez-work

hkny (approved):

PR #78, opened 2022-09-29
feat: update all things SCHEMA to Schema

morrieinmaas:
also update the example schema JSON

Signed-off-by: Moriarty <moritz@animo.id>


spec/data_flow_setup.md
```diff
@@ -73,57 +73,46 @@ presentation of proofs.
...
-        "birthdate",
-        "firstname",
-        "uuid"
-    ],
-    "name": "BasicIdentity",
-    "version": "1.0.0"
+    "id": "https://www.did.example/schema.json",
+    "name": "Example schema",
+    "version": "0.0.1",
+    "attr_names": ["name", "age", "vmax"]
```
morrieinmaas, line 96: Should this be snake or camel case? - on ledger it's currently snake case whilst indy-shared-rs transforms this to camel case.

spec/data_flow_setup.md
```diff
@@ -73,57 +73,46 @@ presentation of proofs.
...
 }
 ```
 
-* `attr_names` is the array of attribute names (claim names) that will
-  constitute the AnonCred credential of this type.
-* `name` is a string defined by the schema publisher, the name of the schema.
-* `version` is a string defined by the schema publisher that the version of the
-    schema in [semver](https://semver.org/) format. The three part, period (".")
-    separated format MAY be enforced.
+- `id` - An identifier that uniquely identifies the Schema. MUST be a valid URI.
```
TimoGlastra, line 100: I wrote this up incorrectly before, but because of the legacy indy structure it can be something else than an URI. This is how I worded it for the credential definition id:

```suggestion
- `id` - (string) The identifier of the [[ref: Schema]]. The format of the identifier is dependent on the [[ref: AnonCreds Objects Method]] used in publishing the [[ref: Schema]].
```

TimoGlastra (approved):

victormartinez-work (approved):

spec/terminology.md
```diff
@@ -73,13 +73,13 @@ The correlatability of credentials due to the necessity of revealing a persisten
...
+[[def: Schema]]
+~ A Schema object is a template that defines a set of attributes (also known as attribute names or [[ref: claims]]) which are going to be used by [[ref: issuers]] for issuance of [Verifiable Credentials](https://www.w3.org/TR/vc-data-model/) within a Hyperledger Indy network. Schemas have a name, version and can be [written](https://hyperledger-indy.readthedocs.io/projects/node/en/latest/transactions.html#schema) to the ledger by any entity with proper permissions. Schemas can be [read](https://hyperledger-indy.readthedocs.io/projects/node/en/latest/requests.html#get-schema) from a Hyperledger Indy Node by any client.
```
swcurran, line 77: Should remove Indy references from this paragraph.

PR #81, opened 2022-09-29
doc: Cred key correctness proof and blinding hidden values

whalelephant:
- added how libursa currently manages the blinding of all hidden attributes (including link secret)

I am aware that only 1 hidden attribute, link_secret, is used, however, it is possible to have more. 
Right now unsure as to how multiple hidden attributes are useful.

### DIFF
--- a/spec/data_flow_issuance.md
+++ b/spec/data_flow_issuance.md
@@ -71,8 +71,42 @@ A [[ref:Credential Request]] is a formal request from a [[ref:holder]] to an [[r
 
 In order to be able as a [[ref:holder]] to express within a [[ref:Credential Request]] to the [[ref:issuer]] which kind of credential the [[ref:issuer]] shall issue to the [[ref:holder]], the [[ref:holder]] requires the [[ref:CRED_DEF]] from the [[ref:Verifiable Data Registry]] if not already available in local storage (step 6 + 7). The [[ref:Credential Request]] has to reference the same [[ref:CRED_DEF]] and [[ref:nonce]] as given in the preceding [[ref:Credential Offer]]. Besides the [[ref:CRED_DEF]], the [[ref:holder]] also requires his [[ref:link secret]] in a blinded form, as well as the corresponding [[ref: Correctness Proof]] of his [[ref:link secret]]. The [[ref: holder]] has now all relevant data for creating the [[ref:Credential Request]] (step 8).
 
+#### Blinding hidden attributes
+
+The [[ref: link secret]] is a default hidden attribute.
+The blinding process is done specifically for an issuer's specific [[ref:CRED_DEF]].
+The `blinded_secrets` sent to the issuer includes:
+
+*- the aggregated values of all blinded hidden attributes
+*- the hidden attrbute key values
+*- the [[ref:Correctness Proof]]
+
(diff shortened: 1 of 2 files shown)


whalelephant:
Converted PR to draft as adding Credential Key Correctness Proof so to complete #todo-12

spec/data_flow_issuance.md
```diff
@@ -71,8 +71,42 @@ A [[ref:Credential Request]] is a formal request from a [[ref:holder]] to an [[r
 
 In order to be able as a [[ref:holder]] to express within a [[ref:Credential Request]] to the [[ref:issuer]] which kind of credential the [[ref:issuer]] shall issue to the [[ref:holder]], the [[ref:holder]] requires the [[ref:CRED_DEF]] from the [[ref:Verifiable Data Registry]] if not already available in local storage (step 6 + 7). The [[ref:Credential Request]] has to reference the same [[ref:CRED_DEF]] and [[ref:nonce]] as given in the preceding [[ref:Credential Offer]]. Besides the [[ref:CRED_DEF]], the [[ref:holder]] also requires his [[ref:link secret]] in a blinded form, as well as the corresponding [[ref: Correctness Proof]] of his [[ref:link secret]]. The [[ref: holder]] has now all relevant data for creating the [[ref:Credential Request]] (step 8).
 
+#### Blinding hidden attributes
+
+The [[ref: link secret]] is a default hidden attribute.
+The blinding process is done specifically for an issuer's specific [[ref:CRED_DEF]].
```
ArPhil: Does this mean, that blinding the link secret by the holder always results in the same blinded value for the issuer / creator of the CRED_DEF?

E.g.: Whenever I (as a holder) blind a link secret in order to get a "Ticket Credential" from the same CRED_DEF (again), is my blinded link secret the same value?

whalelephant: Hi @ArPhil, 

No, this is because every time it is blinded, a blinding factory is used and is generated by the holder 
https://github.com/AnonCreds-WG/anoncreds-spec/pull/81/files#diff-880aaf0a262c52811b2c1e2a63b095c92be4ab57432c017c07e9dabe87778bedR133

ArPhil: Thanks @whalelephant, I was not sure (anymore) because of the phrasing here: [https://github.com/AnonCreds-WG/anoncreds-spec/pull/81/files#diff-880aaf0a262c52811b2c1e2a63b095c92be4ab57432c017c07e9dabe87778bedR76](https://github.com/AnonCreds-WG/anoncreds-spec/pull/81/files#diff-880aaf0a262c52811b2c1e2a63b095c92be4ab57432c017c07e9dabe87778bedR76):

`The blinding process is done specifically for an issuer's specific [[ref:CRED_DEF]].`

Sound to me like it is really "CRED_DEF" specific, but as I understood, it is rather "Issuance specific", since for every issuance process for the holder a new blindung factory is generated by the holder.

Not sure if I am the only one who got confused. If not, maybe we can rephrase this part?

whalelephant: yes sure! please comment on anything else that doesn't make sense / confusing so I can rewrite them as well.

PR #82, opened 2022-09-30
feat: changes to revocation registry definition

morrieinmaas:
Signed-off-by: Moriarty <moritz@animo.id>


spec/data_flow_setup.md
```diff
@@ -418,29 +418,21 @@ interested in understanding what use cases there are for `ISSUANCE_ON_DEMAND`.
...
-    "maxCredNum": 1024,
-    "publicKeys": {
-      "accumKey": {
-        "z": "1 0BB...386"
-      }
-    },
-    "tailsHash": "BrCqQS487HcdLeihGwnk65nWwavKYfrhSrMaUpYGvouH",
-    "tailsLocation": "https://api.portal.streetcred.id/agent/tails/BrCqQS487HcdLeihGwnk65nWwavKYfrhSrMaUpYGvouH"
-  }
+    "type": "CL_ACCUM",
```
swcurran, line 429: Need to discuss if we want to change the data structure as much as that -- e.g. `type` vs. `revocDefType` and `cred_def_id`.  Doesn't really get us far, and may generate a bunch of extra code changes that creates risk.

The `publicKeys` cannot be dropped, so needs to be added back.  Again -- keeping the data structure the same will reduce the code impact, since we are dealing with existing code.

spec/data_flow_setup.md
```diff
@@ -450,23 +442,15 @@ The items within the data model are as follows:
...
-the [[ref: issuer]]. The `z` [[ref: REV_REG]] accumulator public key is
-generated using (TODO: fill in details) algorithm. The use of the accumulator
-public key is discussed in the Credential Issuance section, when the publication
-of revocations is described. The calculation of the tailsHash is described in
+* `type` - the type of revocation registry
+* `cred_def_id` - The credential definition id
+* `tag` - the tag of the credential definition 
+* `issuanceType` - the issuance type; either of "ISSUANCE_BY_DEFAULT" or "ISSUANCE_ON_DEMAND"
+* `maxCredNumber` - The maximum amount of Credentials that can be revoked in the Revocation Registry before a new one needs to be started
+* `tailsLocation` - The URL pointing to the related tails file
```
swcurran, line 450: Can't drop the public key and `z` value -- necessary for the crypto.

swcurran (changes requested):
We'll need to discuss the data structure changes with the devs likely to do the work.  It means that the Indy Method will have to marshal the values from the old to new data structures and then the AnonCreds code will have to be (correctly!) updated to use the new values rather than the old -- all extra work.

Also the dropping of the publicKey is not possible -- it is a necessary data value.

swcurran (changes requested):
We'll need to discuss the data structure changes with the devs likely to do the work.  It means that the Indy Method will have to marshal the values from the old to new data structures and then the AnonCreds code will have to be (correctly!) updated to use the new values rather than the old -- all extra work.

Also the dropping of the publicKey is not possible -- it is a necessary data value.

morrieinmaas:
Thanks for the feedback @swcurran - I'll make amends as much as possible. Please just keep me updated about the status of the quotes reply below.

> We'll need to discuss the data structure changes with the devs likely to do the work. It means that the Indy Method will have to marshal the values from the old to new data structures and then the AnonCreds code will have to be (correctly!) updated to use the new values rather than the old -- all extra work.
> 
> Also the dropping of the publicKey is not possible -- it is a necessary data value.

spec/data_flow_setup.md
TimoGlastra: This is the data type that is used by the AnonCreds library (currently indy-shared-rs and indy-sdk) . When you parse the transaction, indy-sdk / vdr will transform it into the other data model.

TimoGlastra:
This PR stills needs to be updated with learning from RWOT

morrieinmaas:
@TimoGlastra @swcurran Anymore feedback on this?

swcurran (approved):
LGTM

spec/data_flow_setup.md
```diff
@@ -407,29 +407,26 @@ interested in understanding what use cases there are for `ISSUANCE_ON_DEMAND`.
...
-  "id": "Gs6cQcvrtWoZKsbBhD3dQJ:4:Gs6cQcvrtWoZKsbBhD3dQJ:3:CL:140384:mctc:CL_ACCUM:1-1024",
-  "revocDefType": "CL_ACCUM",
-  "tag": "1-1024",
-  "value": {
-    "issuanceType": "ISSUANCE_BY_DEFAULT",
-    "maxCredNum": 1024,
+    "type": "CL_ACCUM",
+    "cred_def_id": "Ehx3RZSV38pn3MYvxtHhbQ:3:CL:264697:default",
+    "tag": "MyCustomCredentialDefinition",
+    "issuanceType": "ISSUANCE_ON_DEMAND",
```
TimoGlastra, line 421: Issuance type can be removed I think. If we define the revocation registry state as a bit array, the issuancetype doesn't matter. We always get the revocation state. It only matters for the anoncreds specific implementations that do support this type (and is therefore only needed on the indy ledger, not the data model).

spec/data_flow_setup.md
```diff
@@ -407,29 +407,26 @@ interested in understanding what use cases there are for `ISSUANCE_ON_DEMAND`.
...
 ``` json
 {
-  "credDefId": "Gs6cQcvrtWoZKsbBhD3dQJ:3:CL:140384:mctc",
-  "id": "Gs6cQcvrtWoZKsbBhD3dQJ:4:Gs6cQcvrtWoZKsbBhD3dQJ:3:CL:140384:mctc:CL_ACCUM:1-1024",
-  "revocDefType": "CL_ACCUM",
-  "tag": "1-1024",
-  "value": {
-    "issuanceType": "ISSUANCE_BY_DEFAULT",
-    "maxCredNum": 1024,
+    "type": "CL_ACCUM",
```
TimoGlastra, line 418: This is missing the `id` property

spec/data_flow_setup.md
```diff
@@ -439,23 +436,16 @@ The items within the data model are as follows:
...
-As noted, most of the items come directly from the input parameters provided by
-the [[ref: issuer]]. The `z` [[ref: REV_REG]] accumulator public key is
-generated using (TODO: fill in details) algorithm. The use of the accumulator
-public key is discussed in the Credential Issuance section, when the publication
-of revocations is described. The calculation of the tailsHash is described in
+* `type` - the type of revocation registry
+* `cred_def_id` - The credential definition id
+* `tag` - the tag of the credential definition 
+* `issuanceType` - the issuance type; either of "ISSUANCE_BY_DEFAULT" or "ISSUANCE_ON_DEMAND"
+* `z` - a public key used to sign the accumulator (described further below)
```
TimoGlastra, line 443: I think it would be nice to keep make this a nested list, so it follows the structure of the json

```
* publicKeys - describe public keys
    * accumKey - describe accumKey
        * z - describe z
```

spec/data_flow_setup.md
```diff
@@ -439,23 +436,16 @@ The items within the data model are as follows:
...
-  as described in the [next section](#tails-file-and-tails-file-generation) on
-  [[ref: TAILS_FILE]] generation.
-* `tailsFileLocation`, the input parameter `tailsLocation`
-
-As noted, most of the items come directly from the input parameters provided by
-the [[ref: issuer]]. The `z` [[ref: REV_REG]] accumulator public key is
-generated using (TODO: fill in details) algorithm. The use of the accumulator
-public key is discussed in the Credential Issuance section, when the publication
-of revocations is described. The calculation of the tailsHash is described in
+* `type` - the type of revocation registry
```
TimoGlastra, line 439: `type` is always `CL_ACCUM` at the moment, would be good to document this

spec/data_flow_setup.md
```diff
@@ -439,23 +436,16 @@ The items within the data model are as follows:
...
+* `cred_def_id` - The credential definition id
```
TimoGlastra, line 440: Would be good to include references here, e.g. the `schema_id` property in the cred def data model is as follows:

```
The ID of the [[ref:Schema]] on which the [[ref: Credential Definition]] for the offered [[ref:Credential]] is based.

```

Maybe it can be something like:

```
The id of the [[ref: Credential Definition]] on which the [[ref: Revocation Registry]] is based.
```

spec/data_flow_setup.md
```diff
@@ -488,7 +478,7 @@ Once generated, the array of primes is static, regardless of credential issuance
 or revocation events. Once generated, the SHA256 (TO BE VERIFIED) hash of the
 array of primes is calculated and returned to be inserted into the `tailsHash`
 item of the [[ref: REV_REG]] object (as described in the [previous
-section](#rev_reg_def-object-generation)). Typically, the array is streamed into a
+section](#revocation-registry-definition-object-generation)). Typically, the array is streamed into a
 file (hence, the term "Tails File") and published to a [[def: URL]] indicated by
 the `tailsLocation` input parameter provided by the [[ref: issuer]].
```
TimoGlastra, line 483: I think we should also update  all references of `PRIVATE_REV_REG` to `Private Revocation Registry Definition` or something like that. To be consistent in the naming, and with the cred def changes

spec/data_flow_setup.md
```diff
@@ -601,7 +591,7 @@ THe following is an example of an initial, published [[ref: REV_REG_ENTRY]] obje
 The items in the data model are:
 
 * `revocDefTyep`: the input parameter `type`
```
TimoGlastra, line 593:
```suggestion
* `revocDefType`: the input parameter `type`
```

spec/data_flow_setup.md
```diff
@@ -601,7 +591,7 @@ THe following is an example of an initial, published [[ref: REV_REG_ENTRY]] obje
...
-* `revocRegDefId`: the identifier of the associated [[ref: REV_REG_DEF]]. The
+* `revocRegDefId`: the identifier of the associated [[ref: Revocation Registry Definition]]. The
```
TimoGlastra, line 594: Should we make this more consistent with the revoc reg def?

spec/data_flow_setup.md
TimoGlastra, line 440: Does it make sense to have the `cred_def_id` snake_case and all other properties camelCase?

TimoGlastra (changes requested):

morrieinmaas:
thx will address today

spec/data_flow_setup.md
morrieinmaas: Was wondering the same. Any opinion on that. That was how it (partially was - see: `* `credDefId`: the input parameter `cred_def_id``, old)

spec/data_flow_setup.md
morrieinmaas: Sorry not sure what you mean

spec/data_flow_setup.md
```diff
@@ -407,52 +407,48 @@ interested in understanding what use cases there are for `ISSUANCE_ON_DEMAND`.
...
-  "credDefId": "Gs6cQcvrtWoZKsbBhD3dQJ:3:CL:140384:mctc",
-  "id": "Gs6cQcvrtWoZKsbBhD3dQJ:4:Gs6cQcvrtWoZKsbBhD3dQJ:3:CL:140384:mctc:CL_ACCUM:1-1024",
-  "revocDefType": "CL_ACCUM",
-  "tag": "1-1024",
-  "value": {
-    "issuanceType": "ISSUANCE_BY_DEFAULT",
-    "maxCredNum": 1024,
+    "type": "CL_ACCUM",
+    "id": "Gs6cQcvrtWoZKsbBhD3dQJ:4:Gs6cQcvrtWoZKsbBhD3dQJ:3:CL:140384:mctc:CL_ACCUM:1-1024",
+    "credDefId": "Ehx3RZSV38pn3MYvxtHhbQ:3:CL:264697:default",
```
swcurran, line 420: I think if we use this "legacy Indy" example, the "id" and "credDefId", need to be related. E.g. the prefix of the "id" is the value of the "credDefId", as that is the way it is in legacy Indy.   We could, of course, remove the values and just but holders in place -- and update the text above about it being an actual example.

spec/data_flow_setup.md
```diff
@@ -407,52 +407,48 @@ interested in understanding what use cases there are for `ISSUANCE_ON_DEMAND`.
...
 Sovrin MainNet and instance of Hyperledger Indy.
 
 ```json
 {
-  "credDefId": "Gs6cQcvrtWoZKsbBhD3dQJ:3:CL:140384:mctc",
-  "id": "Gs6cQcvrtWoZKsbBhD3dQJ:4:Gs6cQcvrtWoZKsbBhD3dQJ:3:CL:140384:mctc:CL_ACCUM:1-1024",
-  "revocDefType": "CL_ACCUM",
-  "tag": "1-1024",
-  "value": {
-    "issuanceType": "ISSUANCE_BY_DEFAULT",
```
swcurran, line 423: While we want to deprecate the "ISSUANCE_ON_DEMAND" feature, I don't think we can simply drop the value from the data structure -- at least not in the 1.0 spec.

swcurran (commented):
While the vast majority of this can be merged, we have to look specifically at the reworking done of the data structures to see if that will work.

spec/data_flow_setup.md
TimoGlastra: We can if we also update the revocation list data model. If you have issuance by default none of the items in the list will have a 0 value by default. If you use `ISSUANCE_ON_DEMAND` all values in the list will have a 1 value to start with.

Internally e.g. indy can use this, but I don't think it's needed in the data model if we always use the full state of the revocation list

swcurran (approved):
LGTM -- thanks!

swcurran:
@TimoGlastra -- can you please look at this again.  Have your requests for changes been resolved?  If not, please work with @morrieinmaas to make that happen.  Thanks!

spec/data_flow_setup.md
TimoGlastra: Doesn't matter, we're going to deprecate this object

TimoGlastra (approved):
LGTM.

 I think we should have a discussion on casing as Stephen said. Indy is VERY inconsistent in camelCase vs snake_case and mixes and matches. I think we're fine with changing some of the ledger data models casing, as those can be easily transformed locally (and thus be backwards compatible).

For the schema and cred def we changed the camelCase properties to snake_case. We should do the same for the revocation registry, OR we should revert those changes to the other models. My opinion is that a few cosmetic changes to the ledger models are not a problem as we're already changing the way revocation models work. However I also see the other side. This is related to this PR: https://github.com/hyperledger/anoncreds-spec/pull/105

Can't make the upcoming AnonCreds meeting due to the later time, but will try to attend the one after that and possibly discuss this in more detail

TimoGlastra:
@morrieinmaas can you address the merge conflicts? Then we can merge

morrieinmaas:
> LGTM.
> 
> I think we should have a discussion on casing as Stephen said. Indy is VERY inconsistent in camelCase vs snake_case and mixes and matches. I think we're fine with changing some of the ledger data models casing, as those can be easily transformed locally (and thus be backwards compatible).
> 
> For the schema and cred def we changed the camelCase properties to snake_case. We should do the same for the revocation registry, OR we should revert those changes to the other models. My opinion is that a few cosmetic changes to the ledger models are not a problem as we're already changing the way revocation models work. However I also see the other side. This is related to this PR: #105
> 
> Can't make the upcoming AnonCreds meeting due to the later time, but will try to attend the one after that and possibly discuss this in more detail

Nice one and agreed. Tricky to do naming but let's make this good choices.

I'll address the merge conflicts in a minute

morrieinmaas:
Not sure how to best go about the not signed off failures here as there seems to be commits not signed off that are not from me.

swcurran:
Hmm...that's weird -- I've never had that issue before.  I'll see if I can check with @ryjones on what can be done.

swcurran:
Sigh -- I hate DCO issues.  @TimoGlastra -- probably best if you follow the rebase instructions here: https://github.com/hyperledger/anoncreds-spec/pull/82/checks.  Perhaps @morrieinmaas can do that -- I'm not sure.

swcurran:
@TimoGlastra @morrieinmaas -- found a way around it. All good now.  Now to fix the merge conflicts in my PR :-)

PR #93, opened 2022-10-31
AnonCreds Overview Diagram in trust triangle layout

ArPhil:
Added new AnonCreds overview layout based on trust triangle diagram. Added proper colors.

image

The diagram can currently be found here and is also referenced in the repo / markdown file as comment: https://docs.google.com/presentation/d/1sUT3AgMd288FG1bI5_WtIvWpmSta5pYbdui8zZ9Z2Tk


hkny:
Really nice diagram. It is much easier to categorize the interactions. Do you think it makes sense to put a read/write/update for the issuer towards VDR and read-only operations for the holder and verifier towards VDR? I like the categorization but not seeing any interaction between the verifier and VDR makes me wonder why I would need a VDR. Maybe it could be as a text outside of the diagram.

ArPhil:
> Really nice diagram. It is much easier to categorize the interactions. Do you think it makes sense to put a read/write/update for the issuer towards VDR and read-only operations for the holder and verifier towards VDR? I like the categorization but not seeing any interaction between the verifier and VDR makes me wonder why I would need a VDR. Maybe it could be as a text outside of the diagram.

I agree and know exactly what you mean. 🤕  In earlier versions of the diagramm I tried to add the information about the AnonCreds objects AND the interaction flows of the objects between different roles in detail. I decided, that both is too much for the reader and is rather confusing than providing clarity and simplicity. I could imagine a second diagram based on the first one, which does not show the object relations and multiplicity but focuses on the read / write permissions and flows. But then I believe we end up quickly in duplication of work, because we explain these things in detail in the appropriate sections later on. Not sure how to deal with this and I still believe the textual description assists pretty well in giving a basic understanding of the diagramm. But: In case you have an idea how to combine both in a simple and concise diagramm, I am very open to it and would be happy to add it . 😃

victormartinez-work:
my 2 cents would be that we stick here to the data model otherwise will become too crowded? and maybe in another diagram, we show the interactions of Anoncreds overlaying the trust triangle. Having said that, I'm not very opinionated on the topic.

spec/introduction.md
```diff
@@ -14,9 +14,10 @@ AnonCreds ZKP verifiable credentials provide capabilities that many see as impor
 
 This version (v1.0) removes any dependence on Hyperledger Indy by removing any requirements related to the storage of the objects used in AnonCreds, whether they be stored remotely on a “verifiable data registry” (including Hyperledger Indy) or in local secure storage.
 
-The following diagram and explanation below give a high level overview of all AnonCreds Data objects, their relations and the owner respectively receiver of each of the data objects. 
+The following diagram and explanation below give a high level overview of all AnonCreds Data objects, their relations and the owner respectively receiver of each of the data objects.
```
victormartinez-work, line 17: high-level

victormartinez-work (approved):

hkny (approved):
I think it is very well done. Let's try to put the dependencies (identity trust triangle) either before or shortly after this layout so it is clear to the reader what interactions happen between which entities.

spec/introduction.md
ArPhil: Fixed.

swcurran (approved):

PR #103, opened 2022-11-09
Correcting file name of logo

swcurran:


TelegramSam (approved):
This is the most US/Canadian PR ever.

swcurran:
US/Canadian/British you mean.  The original file names were "colour" and so I started to change them to "color" and then realized that would be confusing, and Hyperledger was global so "colour" is OK.  And then I missed a spot changing them back.  Sigh...

PR #105, opened 2022-11-15
docs(anoncreds): Updating snake case to camel case [DEV-1879]

Tweeddalex:


swcurran:
@Tweeddalex -- there are conflicts in this PR (perhaps you didn't start with the latest?) and there is no DCO (DCO - Developer Certificate of Origin - https://github.com/apps/dco). Might be easiest to do a new PR -- depending on your GitHub guru-ness.  Sorry to be a pain.

swcurran:
Regards the contents of this PR.  While we can easily change variable names for internal processing, we have to be careful in changing the externally defined JSON items that are likely to be referenced in Aries Frameworks, and especially in existing business logic.

An option is to provide a mapping layer from old and new styles so that either style is supported, but only the old-style is emitted (for now). That enables a transition to emitting old-style objects. 

We need to discuss this in the Working Group before such a transition can be initiated.

TimoGlastra:
My opinion on this is:
- Keep models that are exchanged (cred offer, cred request, etc..) the same. This means exchanges that used to work with 0.1 still work in 1.0.
- Models that are only used locally and with the ledger can have properties changed / removed / added. This is what we're doing with the revocation models etc.. An agent can update this locally and with backwards compatibility.

PR #106, opened 2022-11-20
Update to issuance to add details, all steps, and to-dos

swcurran:


swcurran:
A significant update to the issuance doc to add detail to the flow and to identify places for putting in the cryptographic details.

There are 11 "todo" notes in the text. It would be great to pair with someone that knows either the code or has already gone through the PDF generated from the Latex to fill in the todos. Plan is to do the latter step next on that.

TimoGlastra:
> There are 11 "todo" notes in the text. It would be great to pair with someone that knows either the code or has already gone through the PDF generated from the Latex to fill in the todos. Plan is to do the latter step next on that.

@whalelephant is this something you could help out with?

TimoGlastra (approved):
LGTM. I think there's some qustions I have, but it's probably better to merge and address those as separate issues as they're not directly related to these changes.

whalelephant:
Hi! thanks for tagging, perhaps I can circle back with some details of Credential Signature and Signature Correctness Proof from the code + paper?

swcurran:
@ArPhil -- would love to have you check out the additions and changes in this before merging. Do you have some time for that?

TimoGlastra:
> Hi! thanks for tagging, perhaps I can circle back with some details of Credential Signature and Signature Correctness Proof from the code + paper? 

Yes that would be much appreciated!

spec/data_flow_issuance.md
```diff
@@ -20,102 +20,116 @@ sequenceDiagram
...
 
-where:
+### Credential Request
 
-- `c` can be viewed as the committed value derived from the hash of the concatenated byte values in the process of [creating the Credential Definition](#Issuer-Create-and-Publish-Credential-Definition-Object).
+A [[ref: Credential Request]] is a formal request from a [[ref: holder]] to an [[ref: issuer]] to get a [[ref: credential]] based on the [[ref: Credential Offer]] (and the referenced [[ref: Public Credential Definition]]) sent by the [[ref: issuer]] to the [[ref: holder]].
 
-  $c = H(z || {r_i}  || \~{z} ||\~{r_i})$
+On receipt of the [[ref: Credential Offer]], the [[ref: holder]] retrieves the referenced [[ref: Public Credential Definition]] from a [[ref: Verifiable Data Registry]]. The holder MAY want to retrieve the [[ref: Schema]] referenced in the [[ref: Credential Offer]]
+and verify the consistency between the [[ref: Schema]] and [[ref: Public Credential Definition]].
```
ArPhil, line 115: Should we emphasize, that by consistency we mean that attributes are the same in schema and Credential Definition?

spec/data_flow_issuance.md
```diff
@@ -129,110 +143,212 @@ $$= z^{-c} z^{c} s^{\~{x_z}}\ (Mod\ n)$$
...
 :::
 
-The resulting JSON for a created [[ref:Credential Request]] is shown here:
+Where:
+
+* `u`: *TO BE ADDED*
+* `ur`: is `null` if revocation is not active for the [[ref: Public Credential Definition], and if revocation is active is *TO BE ADDED*
+* `hidden_attributes`: is an array of hidden attributes from the list of [[ref: Public Credential Definition]. For AnonCreds v1.0, it is always a single entry of `master_secret`.
+  * The [[ref: holder]]'s blinded [[ref: link secret]] is a default hidden attribute in AnonCreds, meaning it is not explicitly defined in the [[ref: Schema]] list of attributes but is included in both the [[ref: Public Credential Definition]] and all issued [[ref: credentials]]. Whilst it is cryptographically possible to have multiple hidden attributes, in this version of AnonCreds, only [[ref: link secret]] is used.
+* `committed_attributes`: An empty list of attributes in this version of AnonCreds.
```
ArPhil, line 250: Should we anyway provide an explanation, for what it is intended?

spec/data_flow_issuance.md
```diff
@@ -242,71 +358,189 @@ The [[ref:issuer]] has to transmit the whole credential data to the [[ref:holder
...
+Please note the data attribute "birth_dateint" in the example above. The convention of
+putting a `_dateint` suffix on a credential attribute name is used to indicate
+that the field contains a date in the form of an integer, such as "2022.11.21"
+as the integer "20221121" (the number 20,221,121). By putting the date in that
+form, AnonCreds predicates can be applied to the data, such as proving "older
+than 21" based on date of birth without sharing the date of birth. This
+convention was initially defined
+[here](https://github.com/hyperledger/aries-rfcs/tree/main/concepts/0441-present-proof-best-practices#dates-and-predicates)
+by the [Hyperledger Aries](https://www.hyperledger.org/projects/aries)
+community.
```
ArPhil, line 417: I think we should highlight this section with a box, since applying this is the basis for doing predicates?

ArPhil (approved):
Please find my three minor comments. I noticed many "_TO BE _ADDED__", which I believe is OK, since most of them are related to describing the data structures. Besides that I feel, that the description of the issuance flow is much clearer now, than it was before.

spec/data_flow_issuance.md
```diff
@@ -20,102 +20,116 @@ sequenceDiagram
...
+section](#credential-offer) about the [[ref: Credential Offer]].
 
-The check is done as follows:
+The `key_correctness_proof` verification is as follows:
 
-1. checks that all attributes (except the [[ref: link secret]]) in [[ref:Public Credential Definition]] are included in `xr_cap`
-1. Compute $c'$
-1. If $c' == c$, proof is accepted
+1. Check that all attributes in [[ref: Public Credential Definition]] and `master_secret` (an
+   attribute that will be related to the [[ref: link_secret]]) are included in
```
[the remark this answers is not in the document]
swcurran: In this context it is. It is referring to the data element `master_secret`, and in this case it is the blinded link secret.  I think that is right, hence my usage, but open to suggestion.

spec/data_flow_issuance.md
swcurran: I don't think so.  Arguable for sure, but since in this version the element is not used (always empty), and it's not clear what the structure will look like in the next version, I think we can leave it without an intention.

spec/data_flow_issuance.md
swcurran: Yes -- that what I was trying to do with the `::: note` code block.

Do you know how to put a box around it?  Would be happy to that.  The `note` block will look like this in the rendered spec:

image

swcurran:
@ArPhil -- let me know what you think of my answers and we'll decide on merging or not.  Thanks!

spec/data_flow_issuance.md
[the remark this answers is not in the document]
whalelephant: In the current version of ursalib,  it does not check if the `master_secret` is in the `credential_key_correctness_proof`.
https://github.com/hyperledger/ursa/blob/main/libursa/src/cl/prover.rs#L323 
If we are planning to update that for v1 then please ignore this comment. 

At this stage, the prover would not have blinded the `link_secret` yet due to the fact that it needs to be blinded with the attribute public keys in the credential offer. The `master_secret` type in anoncreds-rs looks like it is the unblinded `link_secret`. It is a randomly generated bigNum of size 256bits. 
https://github.com/hyperledger/anoncreds-rs/blob/main/anoncreds/src/data_types/anoncreds/master_secret.rs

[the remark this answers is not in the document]
ArPhil: Thanks for the clarification. I am trying to distinguish the spec the and the (libursa) implementation. I understand that currently we are also describing how the data elements are named in libursa, since we are on our path of making AnonCreds ledger agnostic. So the reason why we still (sometimes) call it `master_secret` is, because it is called `master_secret` in libursa. But even if the datafield (still) is called `master_secret` in the libursa implementation, shouldn´t we talk about the (blinded) `link_secret`all the time in the spec (and mention what data field is meant)?

I would like to avoid in any case, that readers believe, that a `link_secret` is something different than a `master_secret` (which it is not). As alternative, we could say, that the `master_secret` is the "blinded form" of a `link_secret`, but I believe that this makes things even more complicated.

My suggestion is: Always use `link_secret` in the spec, when we talk about `link_secret`, but explain which data elements (`master_secret`, `blinded_ms`) are meant in this context.

spec/data_flow_issuance.md
ArPhil: Nevermind... for some reason I did not see the `::: note` block.

spec/data_flow_issuance.md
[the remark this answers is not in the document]
swcurran: Regards @whalelephant's comment about master_secret not being included in the correctness proof. It's pretty subtle in the spec. and implementation.  The check you highlight doesn't do what is in the "defined" in the spec.  The spec says that we check that all the attributes are in the array.  The code checks that every entry in `vr_cap` has a key associated with it, bypassing checking if master_secret has such a key. AFAIK, no where are we checking for alignment between the list of attributes in the CRED_DEF and in the correctness proof.

I'd say we should propose a separate PR to update the cryptography parts of these document -- add the missing items and check the existing ones.  That specific line I didn't change -- it was added in a first cryptography addition.

As such -- I think we can go ahead with merging this PR.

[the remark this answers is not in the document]
whalelephant: yep, sounds good. The details in [this HIPE](https://github.com/hyperledger/indy-hipe/tree/main/text/0109-anoncreds-protocol#issuer-proof-of-setup-correctness) definitely does include all attributes in the check as well.

spec/data_flow_issuance.md
swcurran: Clarified.

PR #110, opened 2022-12-07
Another try to enable katex -- minimum change

swcurran:
- Enabling katex in the specification
- Temp change to use awhitehead spec-up until a fix is in spec-up

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,5240 +1,8 @@
 {
   "name": "anoncreds-spec",
   "version": "0.1.0",
-  "lockfileVersion": 2,
+  "lockfileVersion": 1,
   "requires": true,
-  "packages": {
-    "": {
-      "name": "anoncreds-spec",
-      "version": "0.1.0",
-      "license": "Apache 2",
-      "dependencies": {
-        "spec-up": "^0.10.6"
-      },
-      "devDependencies": {
-        "ace-custom-element": "1.6.5",
-        "ajv": "^6.12.5",
-        "mocha": "^8.1.3"
-      }
-    },
-    "node_modules/@traptitech/markdown-it-katex": {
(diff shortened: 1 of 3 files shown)


ArPhil (approved):
@swcurran, I checked out your branch, installed the new dependencies and tested it locally. As far as I can judge, katex is working, since (some) LaTeX is interpreted. But I noticed, that not everything is rendered correctly by katex. 

e.g. in the `Credential Offer` section in the `AnonCreds Issuance Data Flow` I figured out, that the line

`$c = H(z || {r_i}  || \~{z} ||\~{r_i})$` has to be rewritten to ` $c = H(z || {r_i} ||$$\tilde{z}$$||$$\tilde{r_i})$`. The `~` seems not to be interpreted, but the `\tilde` command does. Also for some reason, it does not work when multiple variables are using the `\tilde` command within one single LaTeX expression `$ ... $`. But it works when you split up into multiple LaTeX expressions which results in these weird looking `$$` in the middle of the line. These `$$` are the result of one LaTeX expression ending and another one immediately beginning afterwards. The `Vec< ... >` command also does not seem to be interpreted correctly. Using `\Vec{...}` did not help.

Since this PR is (only) about enabling katex in the first place, I believe it is OK, but wanted to point out and give some hints how to fix / modify the LaTex expressions in the whole spec (or fix it in the spec up package?).

whalelephant:
@ArPhil 

Yep, They are [fixed here](https://github.com/swcurran/anoncreds-spec/compare/enable-katex...whalelephant:anoncreds-spec:belsy/specs-updates?expand=1) along with more content.

I will rebase everything when this gets merged.

TelegramSam (approved):

swcurran:
Merged @whalelephant

PR #121, opened 2023-01-03
update schema model to camelCase

TimoGlastra:
Updates the schema model to use camelCase for the attrNames property. this is because the AnonCreds data models use camelCase by default. 

Model in AnonCreds RS (same as Indy Shared RS): https://github.com/hyperledger/anoncreds-rs/blob/main/anoncreds/src/data_types/anoncreds/schema.rs#L19-L26


swcurran (approved):
LGTM. 

QuestionL Will the Indy AnonCreds Methods have to compensate for this?  Read from the ledger the object, and then (if necessary?) convert it to the AnonCreds format?

PR #122, opened 2023-01-03
update credential definition model to camelCase

TimoGlastra:
Updates the credential definition model to use camelCase for schemaId property. this is because the AnonCreds data models use camelCase by default.

Model in AnonCreds RS (same as Indy Shared RS): https://github.com/hyperledger/anoncreds-rs/blob/main/anoncreds/src/data_types/anoncreds/cred_def.rs#L46-L52

### DIFF
--- a/spec/data_flow_setup.md
+++ b/spec/data_flow_setup.md
@@ -195,7 +195,7 @@ MainNet):
 
 ```json
 {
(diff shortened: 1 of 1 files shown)


swcurran (approved):

swcurran:
Discussed on 2023.01.09 AnonCreds Spec Working Group Meeting -- agreed to merge.

Issue #123, opened 2023-01-03
Consider the need of the `tag` property from credential definition and revocation registry definition models


TimoGlastra:

swcurran:
I think we should keep them.

They are useful in keeping the identifiers differentiated, even if they are not required to be used for that purpose. For example, in Indy, if we drop them, we would have to just generate them to make the IDs unique.

As well, as you note above, they are used by wallets to make a (sort of) human readable display.

Most importantly though, `tag` is in the existing implementation and not needed to be removed to make AnonCreds "ledger agnostic", so I don't see a compelling reason to remove them.

TimoGlastra:
Makes sense.

Do you agree though that it's not a requirement to make them unique as is currently the case for indy? I think the identifier should be unique, and for the indy identifier to be unique the tag needs to be unique (per schema/issuer combi for cred defs, per cred def for rev regs), however it's not needed to make the tag unique I would say.

Reason for this is that it is not possible to enforce. How would I know if there's not any other cred def with the same tag if the tag isn't in the id?

swcurran:
Agreed on the tag uniqueness.  The tag is an issuer generated term, and so it is up to them to make the tag whatever they want.

swcurran:
Discussed at AnonCreds Spec Meeting 2023.01.09 -- agreed to close and keep as is.

PR #124, opened 2023-01-03
update revocation registry model

TimoGlastra:
Updates the revocation registry model to use camelCase and follow the anoncreds implementation.

I've not updated the revocation registry entry related content as that'll be replaced with the revocation list model (in a follow up PR). 

I've also updated the wording of the tag to follow that of the credential definition tag, but we may want to rewrite this after an decision has been made regarding #123 

Model in AnonCreds RS (same as Indy Shared RS): https://github.com/hyperledger/anoncreds-rs/blob/main/anoncreds/src/data_types/anoncreds/rev_reg_def.rs#L72-L100


swcurran (approved):
LGTM

swcurran:
Discussed on 2023.01.09 AnonCreds Spec Working Group Meeting -- likely agreement, but want to look at the dropping of the ISSUANCE_BY_* requirements -- needed in AnonCreds or just in the Indy AnonCreds methods.

whalelephant:
The `issuance_by_*` is not needed technically. It is current used  by 2 functions: 

1. By the Issuer when creating credential with revocation (this should not be a problem as the issuer should know this value)
2. By the prover when creating the presentation, prover first updates the revocation_state (containing witness).

There are 2 ways to update revocation state
1. create new witness (with rev_idx, current revocation status list and `issuance_by_*`)
2. update witness (with old_witness, the revocation status list when old_witness was created, new revocation status list)

Downstream changes to this just means that prover must either provide the old revocation list and old_witness or that they know what variant `issuance_by_*` is.

swcurran:
I think this is the summary:

`issuance_by_*` is needed by the AnonCreds Method, but not by AnonCreds, so it is OK to leave it out of AnonCreds data structures.  By the time the data is passed into AnonCreds for presentation creation, the AnonCreds method must have made sure that the full state of the RevReg has been prepared to be passed in.

An AnonCreds Method that stores deltas (such as Indy) MUST capture from the issuer `issuance_by_*` and MUST provide a way for the holder to get that information.

An AnonCreds Method that stores full state (such as Cheqd) need not worry about `issuance_by_*`.

I think we should talk about this to the spec., in the data flow setup section -- perhaps with a one-liner and link to this section in the presentation generation section?

swcurran:
Agreed to merge -- 2023.01.23 Meeting.

Issue #108, opened 2022-11-28
AnonCreds Revocation List Model


TimoGlastra:
Issue to discuss the data model for the AnonCreds revocation list. This is based on what I've presented in a previous AnonCreds Spec meeting, and is also what we're now targeting for the implementation in AFJ/ACA-Py/AnonCreds-Rs.

Here's the data model:

```json
{
  "revRegId": "<revocRegId",
  "revocationList": [0, 1, 1, 1, 1, 1, 0],
  "currentAccumulator": "<current accumulator value>",
  "timestamp": <timestamp-number>
}
```

- `revRegId` (string) - Identifier of the revocation registry
- `revocationList` (number[]) - Bit array defining the status of the credential in the revocation registry. A value of `1` means the credential is revoked, a value of `0` means the credential is not revoked.
- `currentAccumulator` - The current value of the accumulator
- `timestamp` -  the timestamp at which the accumulator value is valid

This is using camelCase. I know in the spec we've been moving to snake_case, but the current AnonCreds implementation uses camelCase, so maybe it's better to keep it like that. Anyway, snake_case or camelCase the data is the same.

An example:

```json
{
  "revRegId": "4xE68b6S5VRFrKMMG1U95M:4:4xE68b6S5VRFrKMMG1U95M:3:CL:59232:default:CL_ACCUM:4ae1cc6c-f6bd-486c-8057-88f2ce74e960",
[...]

PR #125, opened 2023-01-03
revocation list model

TimoGlastra:
Initial attempt to add the revocation list model to the anoncreds spec. It is based on #108 and replaces revocation registry entries. We should probably add some of the stuff about deltas and revocation registry entries to an implementers guide, and add the transformation from deltas/entries to the revocation status list model to the indy anoncreds method spec (is this already defined somewhere?)


swcurran:
Looks good.  The change to the term "Status List" is one we should confirm with the Working Group -- we'll have it for discussion next Monday (2023.01.09).

swcurran:
Discussed on 2023.01.09 AnonCreds Spec Working Group Meeting -- agree to merge/no objections.

PR #126, opened 2023-01-03
add issuer identifier

### LINKED ISSUE #116: Add an issuerId object to all anoncreds registry objects
Create PR for AnonCreds specification that adds the `issuerId` property to all AnonCreds registry objects and links back to the creator of the anoncreds object. AnonCreds methods must specify how to validate the issuer.

For e.g. indy the issuerId will always be the indy did

TimoGlastra:
Adds an `issuerId` property to all anoncreds ledger objects. The revocation status list is still todo, but it needs #125 so I'll add it in another PR.

For now just added a short section that the anoncreds method must define what the issuerId is, and how you can guarantee only the controller of the identifier can / is allowed to publish objects with an issuerId they control.

Fixes #116


spec/anoncreds_methods.md
```diff
@@ -28,14 +28,26 @@ AnonCreds Resolver is possible, as is a Universal AnonCreds Registrar.
...
+
+All AnonCreds objects that are published to a [[ref: Verifiable Data Registry]] ([[ref: Schema]], [[ref: Credential Definition]], [[ref: Revocation Registry Definition]], and [[ref: Revocation Status List]]) contain an `issuerId` reference. This identifier references the creator ([[ref: issuer]]) of the specific AnonCreds object.
+
+AnonCreds identifiers MUST be a Uniform Resource Identifier (URI)
+conformant with [RFC3986](https://www.rfc-editor.org/rfc/rfc3986), although
+one notable exception is permitted. The exception is that for backwards
+compatibility, the Indy DIDs used in the early (pre `did:indy`) open source
+Hyperledger Indy implementation are permitted.
+
+It is not required for an issuer identifier to be resolvable, however AnonCreds methods MUST define what an issuer identifier must look like, and how it can be verified that the creator of AnonCreds objects are the controller of the issuer identifier. It MUST NOT be possible to publish an AnonCreds object with an issuer identifier that the publisher is not authorized to use.
```
swcurran, line 49: I tripped over this paragraph.  Perhaps reword to something like this:

> An AnonCreds object issuer identifier MAY be publicly resolvable. AnonCreds methods MUST define the format of the issuer identifier, and MUST verify that the publisher of the AnonCreds object controls the issuer identifier before publishing the object.

spec/data_flow_setup.md
```diff
@@ -91,12 +91,14 @@ AnonCreds credential of this type. The following is an example [[ref: Schema]]:
 
 ```json
 {
+  "issuerId": "https://example.org",
```
swcurran, line 94: Is this a good example?  A better example would be a DID, even `did:web`.

spec/data_flow_setup.md
```diff
@@ -433,6 +438,7 @@ Sovrin MainNet and instance of Hyperledger Indy.
 
 ```json
 {
+  "issuerId": "Gs6cQcvrtWoZKsbBhD3dQJ",
```
swcurran, line 441: Again -- not a good example.

spec/data_flow_setup.md
TimoGlastra: I would like to use one example other than dids. I can make one `did:web` and another `https://` maybe?

swcurran: I think an https:// example would be OK, but not a domain name with no path.  That's the confusing part. E.g., if you are thinking of a Cheqd URL to reference a resource, use one of those -- e.g. with a unique component.

PR #128, opened 2023-01-03
Update presentation request section, including removing WQL

swcurran:


swcurran:
Need to add <attr>::<attribute>::marker.

swcurran:
Request that we add a second version of the "_did" properties to match the ID used in the data structures (schema, credDef) for non-Indy implementations (and Indy implementations migrate to the new name).

spec/data_flow_presentation_create_request.md
```diff
@@ -1,13 +1,19 @@
...
-step 1 of the [AnonCreds Presentation Data Flow](#anoncreds-presentation-data-flow)
-by creating and sending a presentation request to the Holder.
+The Verifier starts the process in step 1 of the [AnonCreds Presentation Data
+Flow](#anoncreds-presentation-data-flow) by creating and sending a presentation
+request to the Holder.
 
-The presentation request provides information about the required attributes and predicates to be revealed
-by the Holder as well as information about the credential revocation status and other restrictions.
+The presentation request provides information about the attributes and
+predicates the verifier is asking the the Holder to reveal, restrictions on what
```
wip-abramson, line 8: Suggest consitency. If capital for Holder, then capital for Verifier

wip-abramson:
```suggestion
predicates the Verifier is asking the the Holder to reveal, restrictions on what
```

spec/data_flow_presentation_create_request.md
```diff
@@ -1,13 +1,19 @@
...
+verifiable credentials can be the sources for the attributes and predicates, and
```
wip-abramson, line 9: wondering if this should be anonymous or cryptographic credentials to avoid confusion with the W3C Verifiable Credentials.

spec/data_flow_presentation_create_request.md
```diff
@@ -1,13 +1,19 @@
...
+limitations on the freshness of the credential revocation status. Presentation
+requests are at the "business logic" layer, and do not have any cryptographic
```
wip-abramson, line 11: Not sure what is meant by "do not have any cryptographic component". I would say the nonce is a vital cryptographic component of the request

spec/data_flow_presentation_create_request.md
```diff
@@ -27,43 +33,61 @@ The presentation request is generated by the Verifier in JSON format:
...
-  presents a decimal number (use `anoncreds_generate_nonce` function to generate a correct nonce as 80-bit number)
+* `name` is a string set by the verifier, a name for the presentation request.
+* `version` is a string set by the verifier, the version of the presentation request
+* `nonce` is a string, a decimal, 80-bit number generated by the verifier that
+  SHOULD be unique per presentation request. The nonce is used
+  to prevent replay attacks is used in creating and verifying the presentation.
 * `requested_attributes` specify the set of requested attributes
-  * `attr_referent` is a proof-request local identifier for the requested attribute.
-  * `attr_info` describes a requested attribute, see [attr_info](#attr_info)
+  * `attr_referent` is a verifier-defined identifier for the requested attribute(s).
```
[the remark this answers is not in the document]
wip-abramson: Or actually maybe not hmmm. Depends if the schema type is a restriction on the request I suppse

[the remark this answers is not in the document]
wip-abramson: Ah I remember actually. The attr_referent is literatlly anything the verifier wishes it to be. Its the info that has the schema relevant stuff

spec/data_flow_presentation_create_request.md
```diff
@@ -27,43 +33,61 @@ The presentation request is generated by the Verifier in JSON format:
...
 ```json
 {
-    "name": Optional<string>,
-    "names": Optional<[string, string]>,
-    "restrictions": Optional<wql query>,
-    "non_revoked": Optional<non_revoc_interval>,
+    "name": <string>,
+    "names": <[string, string]>,
+    "restrictions": <restrictions>,
+    "non_revoked": <non_revoc_interval>,
```
wip-abramson, line 62: These should be defined in more detail as above. Especially restrictions

spec/data_flow_presentation_create_request.md
```diff
@@ -72,24 +96,92 @@ The presentation request is generated by the Verifier in JSON format:
...
-* `p_type` is a string, the type of the predicate, with possible types [">=", ">", "<=", "<"]
-* `p_value` is an integer value, the value of the predicate
-* `restrictions` is an (optional) Json - Wql query filter for credentials searching, see [wql_query](#wql_query)
-* `non_revoked` specifies an optional non-revocation interval `non_revoc_interval`, see [Request Non-Revocation Proofs](#request-non-revocation-proofs) section.
-  If specified, the Holder must proof non-revocation for date in this interval for this predicate, this interval overrides the
-  proof level interval.
+* `name` (required) is a string, the name of the attribute from a source
+  credential to use in the predicate expression.
+  * The name is case insensitive and spaces are ignored.
+  * To be useful, the attribute in the source credential MUST be an integer, but
```
wip-abramson, line 107:
```suggestion
  * To be useful, the attribute value in the source credential MUST be an integer, but
```

I think we need to be careful about distinguishing between attribute and attribute value. Or perhaps it is attribute name and attribute. Not sure the best approach, but I initially read this and thought it refered to the name of the attribute being an integer.

spec/data_flow_presentation_create_request.md
swcurran: Agreed. In general, we are going with lower case for "verifier", etc. throughout, and all instances should be `[[ref: <term>]]`, e.g. `[[ref: verifier]]`. I'll do a pass to verify consistency on this PR.

spec/data_flow_presentation_create_request.md
swcurran: Everything in this spec is about AnonCreds, so I think it should be clear.  However, your comment will become more fun when we add the part about generating AnonCreds credentials in W3C VC Data Model Standard format, and AnonCreds presentations in W3C VP Data Model Standard format.  :-).  It will be a pretty easy addition to the spec.

spec/data_flow_presentation_create_request.md
swcurran: Good point.  In "real implementations" the business logic submits the core of the presentation request, and the Aries Framework decorates it as needed.  "As needed" in the case of the presentation request is pretty much only the nonce. But you are right, it is an important component.

spec/data_flow_presentation_create_request.md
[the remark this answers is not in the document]
swcurran: Yes - your last comment is right.

spec/data_flow_presentation_create_request.md
swcurran: Good point on the "attribute name" and "attribute value".  I was struggling a bit with that in writing this. I'll do another pass and see if I can make it clearer.

spec/data_flow_presentation_create_request.md
swcurran: Addressed with latest commit.

spec/data_flow_presentation_create_request.md
swcurran: Addressed with latest commit.

spec/data_flow_presentation_create_request.md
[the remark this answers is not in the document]
swcurran: No change made to spec.  I don't think it is needed.

spec/data_flow_presentation_create_request.md
swcurran: Adjusted the text to address this.  Increasing the confusion is the use of "attribute" for both the requested attributes in the presentation requests, and the attribute names (and values) from the source credentials used to satisfy the presentation request.  I added a paragraph to point out the double use to the specification reader.

We could change all of the "attribute" references to the source credentials to be "claims". However that would leave some remaining confusion because of the naming used in the JSON -- e.g. "attr".  I didn't do that, but request that others weigh in what they think -- @wip-abramson @TimoGlastra @whalelephant and others.

spec/data_flow_presentation_create_request.md
```diff
@@ -72,24 +102,103 @@ The presentation request is generated by the Verifier in JSON format:
...
+credentials that satisfy the `restrictions` expression for each
+attribute/predicate entry. Each element of the logic expression is a property of
+source credentials and a value to be matched for that property. The following
+properties can be specified in the JSON. All except the `marker` property is
+specified with a value that must match the property. For the `marker` property,
+the value is always `1`.
+
+* `schema_id` - the identifier of the schema upon which the source credential is
+  based.
+* `schema_issuer_did` - the identifier, usually a [[ref: DID]], of the publisher
```
TimoGlastra, line 146: This should also include `schema_issuer_id` and `issuer_id`, and mark that the `_did` variants are deprecated. Can also do it in a separate PR if you'd like

TimoGlastra (approved):

Issue #107, opened 2022-11-25
Determine what to do with the prover_did property


TimoGlastra:
It is not required to use AnonCreds with dids. And as I understand it the prover_did doesn't actually have to be a did. It's used internally by ursa as a context string. So I think it just needs to be unique.

@whalelephant you know more about the exact working of this :)

My suggestion would be to rename it to something else that represents a unique identifier, but doesn't necessarily have to be a did. One downside of changing this is that's it is sent to the issuer, meaning it will break backwards compability.

We could also deprecate but allow it, and add a new property that doesn't have to be a did. this way we keep backwards compat. Or we keep the property but allow it to not be a did, but this will break e.g. ACA-Py that currently vaildates whether it is a did.

Thoughts?

whalelephant:
@TimoGlastra 

IIUC, you are right that it does not need to be a DID in Ursa and that the format is actually not checked.

I understand that the `prover_did` is in the credential request that the indy-sdk builds. 
Just a note: this field is not included in the cryptographic material in the request (which only requires blindedSecret + blindedSecretCorrectnessProof),
i.e. this is for the issuer to identify the prover sending the request and the prover's revocation index

In terms of backwards compatibility, is there something ACA-Py does with this field other than validating the format?

swcurran:
I just checked in the ACA-Py Alice-Faber demo and the field is populated by the holder's DID in its connection with the issuer -- which makes no sense at all. The DIDComm DIDs have nothing to do with the issuance of an AnonCreds verifiable credential.  I suspect that when the code was developed, the ACA-Py developer did not know what AnonCreds meant to be put in the field, so they used the only DID the holder had in context. And since it didn't break anything, it was ignored.

In searching ACA-Py for `prover_did`, its referenced in the indy credential request model, and used (badly) in tests.  In some of the tests, it is populated with the issuer's DID.  In the tests it looks like the developer assumed it would be an Indy DID (did:sov), but that can't be the case, as in the general case, the holder does not have a ledger-based (public) DID.

My recommendation is to try to do a bit of archeology into the indy-sdk AnonCreds (and perhaps anoncreds-rs) code and see if the purpose of the field can be determined.  Perhaps it is meant as a unique identifier for the holder that the issuer can hold on to -- but it has lots of such values to use.

My bet is that it is never used in AnonCreds.

[...]

PR #136, opened 2023-01-19
Update information about prover_did in credential request

swcurran:
Addresses issue #107  and matches implementation in [anoncreds-rs#57](https://github.com/hyperledger/anoncreds-rs/pull/57/files).


spec/data_flow_issuance.md
```diff
@@ -171,7 +171,17 @@ Complete the data element descriptions in the following list.
 
 :::
 
-* `prover_did`: *To Be Added*
+* `prover_did`: is a deprecated, optional string that MUST NOT be provided when using other
+  than legacy Indy AnonCreds Methods. Specifically, `prover_did` MUST NOT be
+  provided if the [[ref: Credential Definition]] identifier and the [[ref:
+  Credential Definition]] Issuer identifier are URIs.
```
TimoGlastra, line 177: An indy credential definition id is a valid URI i think, so we should be careful with the wording here

spec/data_flow_issuance.md
```diff
@@ -171,7 +171,17 @@ Complete the data element descriptions in the following list.
...
+  * The `prover_did` serves only as a string in the credential signing process to add entropy.
+  * If not provided, a 22-character random alphanumeric string is generated in its place by the [[ref: issuer]].
```
TimoGlastra, line 179: Is this a requirement? Why 22 characters? Do we want to match a legacy indy did? 

If a requirement maybe we can use MUST.

spec/data_flow_issuance.md
```diff
@@ -170,12 +170,33 @@ Complete the data element descriptions in the following list.
...
+  (`cred_idx`) if the credential is revocable, and the resulting string is hashed
+  to create the `credential_context`, an input to the credential signing
+  process. The `credential_context` is the `m2` item in the issued verifiable
+  credential signature.
+  * Historically in Aries agent implementations, the `prover_did` was populated by
+  the [[ref: holder]] with a [[ref: DID]] they hold, usually the DIDComm
+  peer-to-peer DID shared by the the [[ref: holder]] to the [[ref: issuer]].
+  However, the item is not verified by the [[ref: issuer]] as a DID nor as an identifier
+  for the [[ref: holder]], and as such an random string is sufficient.
+  * The [[ref: holder]] can verify the provide `entropy` value was used by the
```
blu3beri, line 188:
```suggestion
  * The [[ref: holder]] can verify the provided `entropy` value was used by the
```

spec/data_flow_issuance.md
```diff
@@ -170,12 +170,33 @@ Complete the data element descriptions in the following list.
...
+    [[ref: issuer] in generating the signature by combining `entropy` with the
+    `cred_idx` value from the issuer (if the credential is revocable), hashing
```
blu3beri, line 190: Should we specify the hashing algorithm used here?

blu3beri (approved):
Two small comments, otherwise LGTM.

spec/data_flow_issuance.md
```diff
@@ -170,12 +170,33 @@ Complete the data element descriptions in the following list.
...
+    the resulting string and checking that the hash matches `m2` in the
+    credential signature.
 * `cred_def_id`: The ID of the [[ref: Public Credential Definition]] on which the [[ref: Credential]] to be issued will be based.
 * `blinded_ms`: The [[ref: link secret]] in its blinded form. Described in detail in the section [Blinding the Link Secret](#blinding-the-link-secret) (below).
 * `blinded_ms_correctness_proof`: The [[ref: Blinded Secrets Correctness Proof]] of the blinded [[ref: link secret]]. Described in detail in the section [The Blinded Link Secret Correctness Proof](#the-blinded-link-secret-correctness-proof) (below).
 * `nonce`: Used for preventing replay attacks and authentication between protocol steps. *Generation Process to be added*
```
genaris, line 196: Initially I thought this nonce was gone because it's not given as an input for credential request creation in anoncreds-rs (it is generated internally by the library). Now I see both entropy and nonce it seems to me hard to distinguish both. Would be nice to specify requirements for it (currently I see it's a cryptographically secure 80-bit number but not sure if it's enforced to be that way).

spec/data_flow_issuance.md
genaris: I don't see the field `cred_idx` mentioned in other parts of the spec. Are you referring to `cred_rev_id`?

Issue #130, opened 2023-01-09
Validation of anoncreds objects version


blu3beri:
I was curious if there is any validation we have to do, could not find it in the spec, with the anoncreds objects version. Do we allow any string? If we have no intention of validation of versions this can be marked as completed.

TimoGlastra:
What do you mean with validation of the anoncreds version? Where is this present? Do you have an example?

blu3beri:
> What do you mean with validation of the anoncreds version? Where is this present? Do you have an example?

Yeah should have elobaroted on this further, but I am talking about the version inside an anoncreds object like schema: (I thought it was present in more objects, but only schema it seems).

```rust
#[derive(Debug, Clone, Serialize, Deserialize)]
#[serde(rename_all = "camelCase")]
pub struct Schema {
    pub name: String,
    pub version: String, <----
    pub attr_names: AttributeNames,
    pub issuer_id: IssuerId,
}

swcurran:
Discussed on 2023.01.09 AnonCreds Spec Working Group Meeting -- add note in spec. about Indy using semver, but otherwise, it is just a string, and not validated in AnonCreds.

rodolfomiranda:
[PR](https://github.com/hyperledger/anoncreds-spec/pull/142) created

PR #142, opened 2023-02-15
Schema version note

rodolfomiranda:
Adding a note as discussed in #130 .
I've only updated `version` of schema. However `version` and `ver` also appear in Create Presentation Request as:
- `version` is a string set by the [verifier](https://hyperledger.github.io/anoncreds-spec/#term:verifier), the version of the [presentation request](https://hyperledger.github.io/anoncreds-spec/#term:presentation-request) 
- `ver` is an optional string, specifying the [presentation request](https://hyperledger.github.io/anoncreds-spec/#term:presentation-request) version.
If omitted, “1.0” is used by default.
“1.0” to use unqualified identifiers for restrictions
“2.0” to use fully qualified identifiers for restrictions

Having `version` and `ver` in the same object looks weird. Was this already discussed?


swcurran:
Agreed the names are confusing. Is this a good time to change them?

To me, the obvious one to change is the `version` -- the one set by the verifier. We could use `tag` as is used elsewhere for the verifier to apply to a presentation request. They can use a `semver` in the tag if they want -- or anything else.

The concept of `ver` needs to remain, and while I would prefer it be the full word, I recommend that we keep it as is.

Thoughts?

rodolfomiranda:
agree. `tag` is also used in CredDef and RevReg

swcurran:
Could you please resolve the conflicts @rodolfomiranda ?

swcurran (approved):

PR #148, opened 2023-06-19
Added tails file prime generation and tails file hashing

aritroCoder:
Before:
image

After:
image
image


andrewwhitehead:
The entries in the tails file aren't actually primes, they are G2 curve points. The process for generating them looks correct here. Essentially, it is the case that `tail[index] = g_dash * (gamma ** index)`, where `gamma` is a scalar field element. I'm not sure I would describe `g_dash` as 'the' generator of the G2 group as it seems to be picked randomly for each registry.

aritroCoder:
> The entries in the tails file aren't actually primes, they are G2 curve points. The process for generating them looks correct here. Essentially, it is the case that `tail[index] = g_dash * (gamma ** index)`, where `gamma` is a scalar field element. I'm not sure I would describe `g_dash` as 'the' generator of the G2 group as it seems to be picked randomly for each registry. Every curve point is technically a generator but there is also a standard one that is used elsewhere.

image
This text is from another part of the same specification where they call G2 as the generator

aritroCoder:
I have updated the first message with the new updated screenshot of edited spec. Also, I have changed the mention of  `primes` with `points in curve G2` in that section (I guess it needs a separate issue to track down and replace all such instances in case this is present elsewhere too)

spec/data_flow_setup.md
```diff
@@ -444,30 +444,42 @@ The identifier for the [[ref: Revocation Registry]] is dependent on where the
...
 The second of the outcomes from creating of a [[ref: Revocation Registry]] is a [[ref:
 TAILS_FILE]]. The contents of a [[ref: TAILS_FILE]] is an array of calculated
-prime integers, one for each credential in the registry. Thus, if the [[ref:
+points on curve `G2`, one for each credential in the registry. Thus, if the [[ref:
 Revocation Registry]] has a capacity (`maxCredNum`) of 1000, the [[ref: TAILS_FILE]] holds
-an array of 1000 primes. Each credential issued using the [[ref: Revocation Registry]] is
+an array of 1000 `G2` curve points. Each credential issued using the [[ref: Revocation Registry]] is
 given its own index (1 to the capacity of the [[ref: Revocation Registry]]) into the array,
-the index of the prime for that credential. The contents of the [[ref;
+the index of the point for that credential. The contents of the [[ref;
```
swcurran, line 451: Some cleanup of text around what you changed. I’ll add a few of these.

Fix the “;” in the “ref:” at the end of this line.

spec/data_flow_setup.md
```diff
@@ -444,30 +444,42 @@ The identifier for the [[ref: Revocation Registry]] is dependent on where the
...
-::: todo
-To Do: Document hashing of the tails file ([see also](https://github.com/hyperledger/indy-shared-rs/blob/d22373265f7c4cf93d59dd3c111251ef96d6a63d/indy-credx/src/services/tails.rs#L151)).
-:::
+- Use a hasher with any hashing algorithm(`SHA256` in [anoncreds-rs implementation](https://github.com/hyperledger/anoncreds-rs/blob/9c915bb77bc4e033cc6d28d45e330ee5bda26211/src/services/tails.rs#LL148C1-L148C37)).
+- append the tails file version and all the bytes of `G2` curve points one by one into the hasher.
+- Compute the hash digest.
 
-::: todo
-To Do: Document the process for generating the primes.
+The process of generating the primes that populate the [[ref: TAILS_FILE]] are `tail[index] = g_dash * (gamma ** index)`
```
swcurran, line 463: Should the reference to “primes” here be changed as well?

BTW — are “points on the curve G2” prime numbers?

Question: Should we define in the reference section what G1 and G2 are?  Is that needed?  I don’t know — I just sort of gloss over them.  And those that know anything about this stuff probably don’t need an explanation?

spec/data_flow_setup.md
```diff
@@ -444,30 +444,42 @@ The identifier for the [[ref: Revocation Registry]] is dependent on where the
...
+
+::: note
```
swcurran, line 465: Should this be in a note section?  I’m not sure.

Should the actual math expressions in the bullet points be in Latex (as well or instead)?  Does that make it clearer/more useful?

spec/data_flow_setup.md
```diff
@@ -444,30 +444,42 @@ The identifier for the [[ref: Revocation Registry]] is dependent on where the
...
+- To generate a tail prime for an attribute located at a specific index, follow the steps.
+- Convert index into an array of bytes(`u8`) using little endian ordering.
+- Create an element belonging to the finite field group from the `u8` array.
+- Calculate `pow` by doing modular exponentiation of revocation private key(`gamma`) with the finite field element previously calculated.
+- Multiply `pow` by `g_dash`, which is the generator of elliptic curve group `G2`, and this should be the required point of the prime.
+- Convert this tail point to an array of bytes(`u8`), and put them into the file as a slice buffer.
+- Repeat for all the attributes.
+- Close the file buffer.
+
+Relevant links: [Anoncreds-rs repository](https://github.com/hyperledger/anoncreds-rs/blob/9c915bb77bc4e033cc6d28d45e330ee5bda26211/src/services/tails.rs#LL148C1-L148C37), [Ursa repository](https://github.com/hyperledger-archives/ursa/blob/c29fdaa96bbe9ce3ea2beb4d5fbe98ed7c96f867/libursa/src/cl/mod.rs#L514)
```
swcurran, line 477: Please switch the reference to the Ursa repository in favour of the new https://github.com/hyperledger/anoncreds-clsignatures-rs repository.

spec/data_flow_setup.md
```diff
@@ -444,30 +444,42 @@ The identifier for the [[ref: Revocation Registry]] is dependent on where the
...
 :::
 
-Once generated, the array of primes is static, regardless of credential issuance
+Once generated, the array of points is static, regardless of credential issuance
 or revocation events. Once generated, the SHA256 (TO BE VERIFIED) hash of the
```
swcurran, line 481: I think the “(TO BE VERIFIED)” can be removed.

spec/data_flow_setup.md
aritroCoder: fixed

swcurran:
Looks great — nice work.  A few minor cleanups mostly in the text that existed before you started.

Not sure about the “To Do” right below your PR.  Is there something you can add about for the format of the file (e.g. how to extract the points from it), or should it be removed?

spec/data_flow_setup.md
```diff
@@ -444,30 +444,42 @@ The identifier for the [[ref: Revocation Registry]] is dependent on where the
...
-the index of the prime for that credential. The contents of the [[ref;
+the index of the point for that credential. The contents of the [[ref;
 TAILS_FILE]] is needed by the [[ref: issuer]] to publish the current state of
 revocations within the [[ref: Revocation Registry]] and by the [[ref: holder]] to produce
 (if possible) a "proof of non-revocation" to show their issued credential has
 not been revoked.
 
-The process of generating the primes that populate the [[ref: TAILS_FILE]] is as
-follows:
+The process for hashing the [[ref: TAILS_FILE]] is as follows:
```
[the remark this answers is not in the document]
aritroCoder: yeah did it

spec/data_flow_setup.md
aritroCoder: Yes, I missed that line, i'll change it now

I dont think "points on curve G2" are prime. I think there might be some discrepancy between the specification and actual implementation (like it was initially thought to be implemented as prime numbers but later while coding they found out using points could be easier).

G1 and G2 are defined a few sections before in the spec. I am attaching a screenshot of the portion of the spec:

image

But we can add the definitions of G1 and G2 for this context here, that can make this section easier for newcomers

spec/data_flow_setup.md
aritroCoder: I think here we are dealing with how variables are expressed in the actual code rather than in latex (for example we write `g_dash` in the spec instead of writing it in latex). So using latex won't be much useful in my opinion, as it might instead create confusion to the reader.

spec/data_flow_setup.md
aritroCoder: done

spec/data_flow_setup.md
swcurran: Note that definitions can have synonyms, so a single definition would be enough for references to the terms “G1”, “G2” at least, perhaps also “pairing friendly elliptical curve”.

aritroCoder: yeah, so then I guess it's fine as it is

aritroCoder:
> Looks great — nice work. A few minor cleanups mostly in the text that existed before you started.
> 
> Not sure about the “To Do” right below your PR. Is there something you can add about for the format of the file (e.g. how to extract the points from it), or should it be removed?

for the `TO DO` where we have to define the format of the tails file, I am guessing we have to write what all data is written in the file, and in which order.

aritroCoder:
> The values are elliptic curve points and not primes encoded. I need to look deeper as to how those points are encoded

Yeah, that's the reason why the are encoded to the octet stream, which makes it difficult to create a human understandable format

andrewwhitehead:
The points are encoded as uncompressed big-endian coordinates: `{x.0, x.1, y.0, y.1}` for 32*4 = 112 bytes, for the two Fp2 components.

aritroCoder:
I have updated the first message with the new updated screenshot of edited spec after adding the tails file format

spec/data_flow_setup.md
```diff
@@ -486,7 +486,7 @@ section](#revocation-registry-definition-object-generation)). Typically, the arr
 file (hence, the term "Tails File") and published to a [[ref: URL]] indicated by
 the `tailsLocation` input parameter provided by the [[ref: issuer]].
 
-The format of a [[ref: TAILS_FILE]] is as follows(binary points are shown in readable format):
+The format of a [[ref: TAILS_FILE]] is as follows(bytes representing points are shown in readable format):
```
swcurran, line 489: I’m not exactly clear on the format and have a couple of questions that I think might help:

- Likely my ignornance, but I’m not familiar with the “u” notation. I’m guessing it is 2 bytes at the beginning of the file?
- Is the “version:” literal?
- How long (number of bytes) are each of the points?

aritroCoder: - `0u8` is the number 0 as an unsigned 8-bit integer, similiarly for `2u8`
- Version is specifically the version of the file(which is there in the rust implementation [here](https://github.com/hyperledger/anoncreds-rs/blob/9c915bb77bc4e033cc6d28d45e330ee5bda26211/src/services/tails.rs#L153C16-L153C16)) 
- A point is a collection of two integers, so it should be 2*4 bytes(for 2D point), but since `ECP2` implemented points as 3D, so it would be 3*4=12 bytes long

swcurran: Got it.  I would suggest then making the file format just bullet points (vs. fenced) and describe what you have above, something like:

* The version of the tails file, as a 2 unsigned bytes. The current version uses the value is `0u8 2u8`.
* A list of the points, one per credential in the Revocation Registry. Each a collection of two integers (<what is in the PR now> implemented as points in 3 dimensions. Each point is 12 bytes long.

The total size of a Tails File is `2+12*Size of the Revocation Registry`.

aritroCoder: Did it, and updated the screenshots in the first message

spec/data_flow_setup.md
```diff
@@ -444,40 +444,54 @@ The identifier for the [[ref: Revocation Registry]] is dependent on where the
...
+::: note
+Detailed process for tails file generation:
+- Create and open the tails file.
+- To generate a tail point for an attribute located at a specific index, follow the steps.
+- Convert index into an array of bytes(`u8`) using little endian ordering.
+- Create an element belonging to the finite field group from the `u8` array.
+- Calculate `pow` by doing modular exponentiation of revocation private key(`gamma`) with the finite field element previously calculated.
+- Multiply `pow` by `g_dash`, which is the generator of elliptic curve group `G2`, and this should be the required point on the curve.
+- Convert this tail point to an array of bytes(`u8`), and put them into the file as a slice buffer.
+- Repeat for all the attributes from index $1$ to $L$, by calculating $([\gamma], [\gamma^2], [\gamma^3], ...[\gamma^L], [\gamma^{L+2}], [\gamma^{L+3}], ..., [\gamma^{2L}])$. Omit $[\gamma^{L+1}]$ as it can be used to generate fake witness for non-revocation.
```
swcurran, line 468: To be discussed -- I think this is not right. From what I understand:

- 2 bytes for version
- 1...$L are points associated with credentials 1...L
- $L+1 is a duplicate of 1 (filler)
- $L+2...$2L+1 are points associated with credentials 1...L

Hopefully we can clarify this on the July 31, 2023 AnonCreds call.

spec/data_flow_setup.md
```diff
@@ -444,40 +444,54 @@ The identifier for the [[ref: Revocation Registry]] is dependent on where the
...
+
+Relevant links: [Anoncreds-rs repository](https://github.com/hyperledger/anoncreds-rs/blob/9c915bb77bc4e033cc6d28d45e330ee5bda26211/src/services/tails.rs#LL148C1-L148C37), [Anoncreds-CLSignatures repository](https://github.com/hyperledger/anoncreds-clsignatures-rs/blob/f1ae666656054cd73fe765928c0dada64ef21d87/src/mod.rs#L517)
 :::
 
-::: todo
-To Do: Document the process for generating the primes.
-:::
+The process for hashing the [[ref: TAILS_FILE]] is as follows:
+
+- Use a hasher with any hashing algorithm(`SHA256` in [anoncreds-rs implementation](https://github.com/hyperledger/anoncreds-rs/blob/9c915bb77bc4e033cc6d28d45e330ee5bda26211/src/services/tails.rs#LL148C1-L148C37)).
```
swcurran, line 476: I think we need to explicitly declare this is `SHA256` so that others can implement AnonCreds and be compatible with the AnonCreds Rust implementation.

spec/data_flow_setup.md
```diff
@@ -444,40 +444,54 @@ The identifier for the [[ref: Revocation Registry]] is dependent on where the
...
 
 The format of a [[ref: TAILS_FILE]] is as follows:
 
-::: todo
-To Do: Define the format of the Tails File
-:::
+- First two bytes are version number(currently `0u8 2u8`)
+- A list of the points, one per credential in the Revocation Registry. Each point is a collection of three integers implemented as points in 3 dimensions as per `ECP2`. Each point is 3x4 = 12 bytes long.
+
+Thus the total size of a Tails File is 2 + 12*`Size of the Revocation Registry`.
```
swcurran, line 494: I think this should 2+ 12*Size of the Revocation Registry+6 (the L+1 entry).

swcurran: Each point is 256 bits or 32 bytes. Each coordinate is two field elements, so a total of 4 elements, and 128 bytes per point.

spec/data_flow_setup.md
```diff
@@ -444,40 +444,54 @@ The identifier for the [[ref: Revocation Registry]] is dependent on where the
...
 TAILS_FILE]]. The contents of a [[ref: TAILS_FILE]] is an array of calculated
-prime integers, one for each credential in the registry. Thus, if the [[ref:
+points on curve `G2`, one for each credential in the registry. Thus, if the [[ref:
 Revocation Registry]] has a capacity (`maxCredNum`) of 1000, the [[ref: TAILS_FILE]] holds
-an array of 1000 primes. Each credential issued using the [[ref: Revocation Registry]] is
+an array of 1000 `G2` curve points. Each credential issued using the [[ref: Revocation Registry]] is
 given its own index (1 to the capacity of the [[ref: Revocation Registry]]) into the array,
-the index of the prime for that credential. The contents of the [[ref;
+the index of the point for that credential. The contents of the [[ref:
 TAILS_FILE]] is needed by the [[ref: issuer]] to publish the current state of
```
swcurran, line 452: This has changed recently in the implementation. Please remove "by the [[ref: issuer]] to publish the current state of revocations within the [[ref: Revocation Registry]] and"

spec/data_flow_setup.md
swcurran: In the sequence part `$([\gamma], [\gamma^2], [\gamma^3], ...[\gamma^L], [\gamma^{L+2}], [\gamma^{L+3}], ..., [\gamma^{2L}])$` add in `[\gamma]` where L+1 has been left out.

Remove the "Omit..." sentence and replace it with `Note that Instead of inserting $[\gamma^{L+1}]$ in the sequence, insert the value $[\gamma]$ (the first value in the sequence) in its place, and then continue with $[\gamma^{L+2}]$ and on to $[\gamma^{2L}]$. $[\gamma^{L+1}]$ is not used by holders generating the [[def: Non-Revocation Proof]] and a dummy value is inserted in its place.`

spec/data_flow_setup.md
swcurran: Change -- "Use a hasher with any hashing algorithm(`SHA256` in" to "Use the `SHA256` hashing algorithm"

spec/data_flow_setup.md
```diff
@@ -444,40 +444,54 @@ The identifier for the [[ref: Revocation Registry]] is dependent on where the
...
+- append the tails file version and all the bytes of `G2` curve points one by one into the hasher.
+- Compute the hash digest.
 
-Once generated, the array of primes is static, regardless of credential issuance
-or revocation events. Once generated, the SHA256 (TO BE VERIFIED) hash of the
-array of primes is calculated and returned to be inserted into the `tailsHash`
+
+Once generated, the array of points is static, regardless of credential issuance
+or revocation events. Once generated, the SHA256 hash of the
+array of points is calculated and returned to be inserted into the `tailsHash`
```
swcurran, line 483: Change this:  

- Once generated, the SHA256 hash of the array of points is calculated and returned to be inserted into the `tailsHash`

to this:

-  The SHA256 hash of the array of points is returned to be inserted into the `tailsHash`

spec/data_flow_setup.md
swcurran: @andrewwhitehead -- can you confirm my phrasing?

spec/data_flow_setup.md
swcurran: @andrewwhitehead -- can you please confirm the correct size of the file?

spec/data_flow_setup.md
swcurran: @aritroCoder -- confirmed with Andrew -- `It's a copy of index 0, which isn't used otherwise. g_dash*gamma^0 which is just g_dash`

swcurran (approved):

PR #151, opened 2023-07-06
updated notations section with new notations

### LINKED ISSUE #150: Create a comprehensive notation section at the begining of the spec.
Similiar to [BBS+ signatures specification](https://www.ietf.org/archive/id/draft-looker-cfrg-bbs-signatures-01.html#section-1.2), this spec can have a `notations` section that properly covers all the different mathematical and cryptographic symbols to the reader. Most parts of it will remain same as the given example, with some minor additions/deletions.

aritroCoder:
Fixes #150 
Updated the `notations` section in the spec

image

Let me know if anything I missed here


swcurran (approved):
Looks good! Thanks.

PR #158, opened 2023-08-03
Added description of the nonce in the credential request data model and document how the nonce from the Credential Offer is used.

aritroCoder:


spec/data_flow_issuance.md
```diff
@@ -117,6 +117,10 @@ the [[ref: Credential Offer]] and verify the consistency between the list of
 attributes in the [[ref: Schema]] and in the [[ref: Public Credential
 Definition]].
 
+The nonce of the [[ref: Credential Offer]] is used to generate the proof of correctness
+for blinded credential secrets, where it is hashed with the blinded secrets to
```
swcurran, line 121: Quick check -- is that "blinded secret" or "blinded secrets"?  I think there is just one?

aritroCoder: We can technically blind multiple attributes but in the spec they are referred as a single blinded secret

swcurran: Sounds good.

swcurran (approved):

PR #160, opened 2023-08-10
Blinded Link Secret proof model completed

aritroCoder:
I am a bit unsure about the 4th attribute `r_caps` in the structure of the proof , as it is not mentioned in any of anoncreds documents or codebase which I have seen. Please verify and suggest what should I add there.


swcurran (approved):
LGTM

PR #164, opened 2023-08-26
Completed credential signature data model

aritroCoder:
This pull request should be merged only after merging #162, otherwise the changes does not make much sense


spec/data_flow_issuance.md
```diff
@@ -445,20 +445,13 @@ community.
...
-* `e` is the *TO BE ADDED*. It is constructed as follows:
-  * *TO BE ADDED*
-* `v` is the *TO BE ADDED*. It is constructed as follows:
-  * *TO BE ADDED*
+* `m_2` is a linkable identifier to the holder that is also called the `master_secret` in old versions. It is constructed as follows:
+  * $m_2 = H(i || \mathcal{H})$, where $i$ is an index assigned to the holder, and $\mathcal{H}$ is an identifier with which the [[ref: holder]] is known to the [[ref: issuer]].
+* `a` is the signature of the blinded known attributes. It's generation is given above.
+* `e` is a random prime generated by the [[ref: issuer]] for creating signature.
+* `v` is a number generated by the [[ref: holder]] to unblind the signature of the blinded attributes. It is constructed as follows:
+  * $v = v' + v''$, where $v'$ is the blinding factor which the holder has and $v''$ is the random number generated by the issuer.
```
swcurran, line 454: "and $v''$ is the random number generated by the issuer"

I think it should be a "is a random number...", as I assume they generate it on the fly?  I would think that the attributes (e.g. the size) of the number should be indicated?

mikelodder7 (approved):

spec/data_flow_issuance.md
```diff
@@ -445,20 +445,13 @@ community.
...
-  * *TO BE ADDED*
-* `a` is the *TO BE ADDED*. It is constructed as follows:
-  * *TO BE ADDED*
-* `e` is the *TO BE ADDED*. It is constructed as follows:
-  * *TO BE ADDED*
-* `v` is the *TO BE ADDED*. It is constructed as follows:
-  * *TO BE ADDED*
+* `m_2` is a linkable identifier to the holder that is also called the `master_secret` in old versions. It is constructed as follows:
+  * $m_2 = H(i || \mathcal{H})$, where $i$ is an index assigned to the holder, and $\mathcal{H}$ is an identifier with which the [[ref: holder]] is known to the [[ref: issuer]].
+* `a` is the signature of the blinded known attributes. It's generation is given above.
```
mikelodder7, line 451: Add how is encoded. I believe for anoncreds 1.0 its base 10

aritroCoder: yes, it is encoded in decimal

aritroCoder: added this into the spec

swcurran (approved):

PR #169, opened 2023-10-05
Added `eq_proof` and `ge_proofs` for presentation

aritroCoder:


swcurran:
Looks good to me -- but I'm not sure I'm helpful here. @mikelodder7 -- would really help to have you review this.

@aritroCoder -- much (but not all...) of the Latex is not rendering in GitHub, which may be expected, and it will work when rendered by SpecUp.  Are you concerned with that?  I tried a quick check, but couldn't see whether you were doing anything different between when it does and does not render properly in the "Rich Diff" display.

aritroCoder:
> Looks good to me -- but I'm not sure I'm helpful here. @mikelodder7 -- would really help to have you review this.
> 
> @aritroCoder -- much (but not all...) of the Latex is not rendering in GitHub, which may be expected, and it will work when rendered by SpecUp. Are you concerned with that? I tried a quick check, but couldn't see whether you were doing anything different between when it does and does not render properly in the "Rich Diff" display.

I don't think I am doing anything different in any case, all of them are proper markdown latex and should be rendered properly. I guess it is a limitation by github

mikelodder7 (approved):

PR #170, opened 2023-10-07
Added aggregate proof generation and data models

aritroCoder:


spec/data_flow_presentation_create_presentation.md
```diff
@@ -271,13 +271,41 @@ To Do: Add algorithm for generating a `ge_proof` and the data values in the proo
...
+        ]
+      }
+```
 
-To Do: Add algorithm for generating the aggregate proof and the data values in the proof.
+where:
 
-:::
+- `c_hash` is the fiat shamir challenge hash $c_H$,
+- `c_list` is the list of commitments $\mathcal{C}$ that we calculated during the challenge hash generation.
```
swcurran, line 308: Could you add what determines how many entries are in the c_list?  Is it based on the number of attributes in the different credentials?  Something else?

aritroCoder: As much as I understand, c_list size depends on the number of attributes to prove in a credential, and it is generated seperately for each credential

spec/data_flow_presentation_create_presentation.md
```diff
@@ -271,13 +271,41 @@ To Do: Add algorithm for generating a `ge_proof` and the data values in the proo
 The `aggregated_proof` proves that the same [[ref: linked secret]] was used to
 issue all of the source verifiable credentials in the presentation.
 
-The `aggregated_proof` is generated as follows:
+The `aggregated_proof` structure is as follows:
 
-::: todo
+```json
+"aggregated_proof": {
```
mikelodder7, line 277: Instead of this maybe use "aggregated_proof": "base10string"

aritroCoder: This is implemented as structure in the code, so should I do 
`"aggregated_proof": "base10string"`
or should I write out the entire structure like
```json
"aggregated_proof": {
        "c_hash": "base10string",
        "c_list": [ ["base10string"] ]
}
```

swcurran: Looking at here (I wrote my previous comment on the Conversation tab) -- put in the structure, and then something like "Here is an example: " and then what you have here.

Looks like the hash is a base10string, but the c_list is a list of integers (not strings...)?

spec/data_flow_presentation_create_presentation.md
```diff
@@ -379,13 +379,41 @@ ge_proofs: [
...
+        ]
+      }
+```
 
-To Do: Add algorithm for generating the aggregate proof and the data values in the proof.
+where:
 
-:::
+- `c_hash` is the fiat shamir challenge hash $c_H$,
+- `c_list` is the list of commitments $\mathcal{C}$ that we calculated during the challenge hash generation. It's size depends on the number of attributes to prove in a credential, and it is generated seperately for each credential.
```
swcurran, line 416: Perhaps a bit clearer to say -- "The list contains an array entry per credential, and within each entry, an integer per attribute to prove in that credential."

Is it "attribute to prove" or simply "attribute".  What does "to prove" mean?  Does it mean revealed?  I'm not sure, so that should be clarified.

spec/data_flow_presentation_create_presentation.md
aritroCoder: yes, `c_list` is the set of the commitments generated for preparing the proof

spec/data_flow_presentation_create_presentation.md
aritroCoder: it's a commitment to the attribute, which means the exact attribute value is not revealed for the proof to remain zero knowledge.

aritroCoder: C stores the blinded commitments of data attributes of issuer's public revocation key, non revocation credential, and then also stores the blinded commitments of the blinders used to generate those commitments

swcurran (approved):

PR #171, opened 2023-10-08
Added requested proof models

aritroCoder:
These includes self-attested attributes, unrevealed attributes and predicates.

### DIFF
--- a/spec/data_flow_presentation_create_presentation.md
+++ b/spec/data_flow_presentation_create_presentation.md
@@ -485,7 +485,7 @@ and `to` values. How to determine an appropriate [[ref: Revocation Registry
 Entry]] to use is up to the holder, and their use of the capabilities of the
 relevant [[ref: Verifiable Data Registry]].
 
-::: Note
+::: note
 
 When Hyperledger Indy is the [[ref: Verifiable Data Registry]] for the [[ref:
 Revocation Registry Entry]], the state is received from Indy ledger as the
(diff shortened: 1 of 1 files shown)


spec/data_flow_presentation_create_presentation.md
```diff
@@ -343,27 +343,39 @@ An entry for each group `names` request attribute from the presentation request.
 
 **Self Attested Attributes**
 
-::: todo
-
-To Do: Describe self-attested attributes data structure
+This is a set any other data that holder wants to prove to verifier that is not signed by any issuer thus it’s attested only by the holder. This is a set of attributes that the holder is attesting to the verifier. The holder is not claiming that the data is true, only that they are attesting to it.
```
swcurran, line 346: I think we should remove the "prove" as it is too strong.  Perhaps better -- "provide".  Also add a "the" in front of verifier.

spec/data_flow_presentation_create_presentation.md
```diff
@@ -343,27 +343,39 @@ An entry for each group `names` request attribute from the presentation request.
...
+```
 
 **Unrevealed Attributes**
 
-::: todo
-
-To Do: Describe unrevealed attributes data structure
-
-:::
+These are a hashmap of unrevealed attributes with their sub proof index numbers.
```
swcurran, line 356: It should be added that these are attributes that the verifier requested in the presentation request, but the holder has decided not to reveal. It does not include the set of attributes that the verifier did not request at all.

I think that statement is correct, isn't it?

swcurran:
A few changes requested, then it is looking good.  Thanks!

swcurran (approved):

Issue #181, opened 2023-11-13
Question the correctness of holders AnonCreds data objects diagram


wip-abramson:
I am wondering if the following diagram is misleading?

![image](https://github.com/hyperledger/anoncreds-spec/assets/18055112/d455ef31-f771-4e92-abe3-123756353046)

To me, interpretting this seems to imply that for every AnonCred credential that the holder recieves they must store the associated blinded link secret that they used during the issuance protocol.

I do not think that is correct. 

My understanding is that I create a blinding of my link secret that is used for the issuance protocol **BUT** as part of that protocol I remove the blinding from the signature I recieve on my credential. Removing this blinding effectively means the holder stores a valid signature - different from the one the issuer created - over the set of attributes of the credential where one of these attributes is the **unblinded** link secret. Therefore, the blinding factor and the blinded link secret used in this protocol are only relevant for the protocol itself. There is no need to store these after the protocol has been completed.

During a presentation protocol, the holder creates a fresh unlinkable blinding for their link secret and an associated proof that they know the link secret that was signed in the credential. Again, this blinding is only used for the single protocol instance. No need to persist it.

I think I would update the spec text associated with the diagram to reflect some of this. The most important point, to me at least, is that ultimately the holder stores credentials which are **a signature over a set of attributes where one attribute is their link secret**

edeykholt:
While I don't know the answer, I see the class diagram has an association with 1..1 multiplicities on both sides, which is suspicious.

swcurran:
A PR #183 was merged on this that updated the text about this, but the image still needs to be updated, so this issue still needs action.

PR #183, opened 2023-11-14
WIP refactor issuance protocol text

wip-abramson:
This is a start to address the issue I raised in #181. 

Not made it all the way through the spec yet - at about Section 7. I seem to be making a fair few changes here, so comments and feedback welcome as I go.

Also, I recommend we update the following diagrams:

- AnonCreds data objects in the introduction. Suggest: remove blinded link secret from holder entirely
- 8. Issuance Flow. Suggest we add a step before 13 that is something like: Remove blinding factor


wip-abramson:
Getting there...

I am not fully certain how the blinding factor is generated in RSA, so we should check this.

wip-abramson:
Okay, I think the last piece is around `8.4  Receiving a Credential`. 

I need to look into the protocol, but I am pretty certain there should be a step here that transforms the credential signature from the issuer (A,e,v) into (A',e',v') using the blinding factor. Where the credential signature stored by the holder is (A',e',v').

wip-abramson:
So I checked the CL signature code.

This is the line where the prover does the operation to remove the blinding from the credential signature [L635](https://github.com/hyperledger/anoncreds-clsignatures-rs/blob/56fb44364d85b15ca9c33d63bc057e99422a41b6/src/prover.rs#L635). Called from [L237](https://github.com/hyperledger/anoncreds-clsignatures-rs/blob/56fb44364d85b15ca9c33d63bc057e99422a41b6/src/prover.rs#L237) in process_credential_signature.

This line is

`p_cred.v = v_prime.add(&p_cred.v)?;`

Which mutates the V valuse of the credential signature.

swcurran (approved):

wip-abramson:
Cool! Apologies I counldn't make the call

swcurran:
Thanks for the PR and issues, @wip-abramson.  Working on the spec today — continuing the cleanup.

PR #185, opened 2023-11-15
Added section about W3C Verifiable Credential/Presentation representation

Artemkaaas:
Add section describing how AnonCreds credentials and presentations can be represented in the W3C Data Model standard.


spec/w3c_representation.md
```diff
@@ -0,0 +1,540 @@
...
+    "encoding": "auto"
+  },
+  "credentialSubject": {
+    "firstName": "Alice",
+    "lastName": "Jones",
+    "age": "18"
+  },
+  "proof": [
+    {
+      "type": "CLSignature2023",
```
TimoGlastra, line 38: I think we should call this something like "AnonCredsProof2023"

spec/w3c_representation.md
```diff
@@ -0,0 +1,540 @@
...
+```
+
+#### Issuance Date
+
+W3C [Issuance Date](https://www.w3.org/TR/vc-data-model/#issuance-date) section requires including of `issuanceDate`
+property to express the date and time when a credential becomes valid.
+
+In the case of W3C AnonCreds credentials, the `issuanceDate` attribute should contain the time when a credential was
+issued or
+transformed from legacy form, and looks the following:
```
TimoGlastra, line 196: I think we might want to add "or use 0000-01-01T00:00:00Z" or something? That way it's really saying nothing

spec/w3c_representation.md
```diff
@@ -0,0 +1,540 @@
...
+
+The defined [@context](#context) includes a definition for the `CLSignature2023` type describing the format of the proof
+entry:
+
+```
+{
+  ... 
+  "proof": [
+    {
+      "type": "CLSignature2022",
```
TimoGlastra, line 228: should be 2023 and propose we change the name

spec/w3c_representation.md
```diff
@@ -0,0 +1,540 @@
...
+    ```
+        "credentialSubject":{
+            ...
+            "age":[
+                {
+                  "type":"AnonCredsPredicate",
+                  "p_type":">=",
+                  "p_value":18
+                }
+            ]
```
TimoGlastra, line 445: Would it be possible to create a more human natural language way to show this you think? also maybe we can choose more human friendly names for `p_type` and `p_value`? just `predicate` and `value` might work?

spec/w3c_representation.md
swcurran: We discussed this and agreed that we should align with what is proposed in the W3C specification.  If the VC is presented in an VP with an AnonCreds signature, the `issuanceDate` will not be included as a signed element and so it is irrelevant what it is.  If the VC is presented in a VP with another signature, the `issuanceDate` will be as signed by the Issuer.

spec/w3c_representation.md
swcurran: Agreed -- I think we should go with the name @TimoGlastra has suggested.

spec/w3c_representation.md
swcurran: Suggest this conversation be taken to the implementation work being done in https://github.com/hyperledger/anoncreds-rs. This spec. will reflect what is decided there, and at this point, this is the (admittedly, small group) consensus there.  There is still time for an adjustment based on other ideas.

swcurran:
@Artemkaaas -- please address the conflicts. Should be easy :-)

spec/w3c_representation.md
```diff
@@ -0,0 +1,533 @@
...
+
+#### Status
+
+W3C [Status](https://www.w3.org/TR/vc-data-model/#status) section defines an optional capability to include credential
+status information.
+
+In the case of W3C AnonCreds credentials, instead of including `expirationDate` property there is defined another
+[Announced Revocation Data Flow](./data_flow_revocation.md).
+
+In order to satisfy this flow, id of `revocationRegistry` must be included into the `credentialSchema` and revocation
```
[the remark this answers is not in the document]
Artemkaaas: This makes sense. I'll update the document to use `credentialStatus` property for exposing id of revocation registry.

spec/w3c_representation.md
```diff
@@ -130,13 +130,6 @@ schema in order to include the information about AnonCreds related definitions t
 * `type` - `AnonCredsDefinition`
 * `schema` - id of [[ref: Schema]]
 * `definition` - id of [[ref: Credential Definition]]
-* `revocation_registry` - (Optional) id of [[ref: Revocation Registry Definition]]
```
swcurran: Did you mean to remove `revocation_registry` as well here? Or was that not supposed to be there, since it is not in the example?

Artemkaaas: It seems you look outdated version. `revocation_registry` is not there anymore.

PR #188, opened 2023-11-27
Update Trust Triangle Diagram to remove Blinded Link Secret

swcurran:
Update the "Trust Triangle" image to reflect the changes in the text from #183 -- removing the Holder having a blinded link secret as a separate object.  Completes issue #181.

@wip-abramson -- I think that works, right?


wip-abramson (commented):
Yep this looks great

Issue #60, opened 2022-07-10
Rationalize the terminology "def"s and the "ref"s in the rest of the spec


swcurran:
The following is a quick'n'dirty command to find the `ref`s and `def`s in the document as of today.

This issue is to rationalize the data. Pick a point where there are no PRs in process for `terminology.md`, add in the needed definitions, and update the documents to fix the references.  Not that the references are case insensitive, so there is no need to align the case of the refs/defs.

BTW - if you drop the "`-h`" on the `grep` command, the name of the file containing the matched ref/def is shown, making it easier to find the fixes.  Also, I'm not certain if only the first or all of the `ref`s on a single line are included in this output. Dropping the `-o` would help to discover that.

```
~/repos/anoncreds-spec [main] 16:08 $ grep -o -h '\[\[ref: [a-zA-Z_]*\]\]' spec/*.md | sort -u
[[ref: claim]]
[[ref: CLAIM_DEF]]
[[ref: claims]]
[[ref: CRED_DEF]]
[[ref: credential]]
[[ref: holder]]
[[ref: Holder]]
[[ref: holders]]
[[ref: Holders]]
[[ref: issuer]]
[[ref: Issuer]]
[[ref: link_secret]]
[[ref: predicates]]
[[ref: PRIVATE_CRED_DEF]]
[[ref: PRIVATE_REV_REG]]
[[ref: REG_REV]]
[[ref: REV_REG]]
[[ref: REV_REG_DEF]]
[[ref: REV_REG_ENTRY]]
[[ref: SCHEMA]]
[[ref: subject]]
[[ref: TAILS_FILE]]
[[ref: VDR]]
[[ref: verifier]]
[[ref: Verifier]]
~/repos/anoncreds-spec [main] 16:08 $ grep -o -h '\[\[def: [a-zA-Z_]*\]\]' spec/*.md | sort -u
[[def: claim]]
[[def: CRED_DEF]]
[[def: credential]]
[[def: DID]]
[[def: holder]]
[[def: issuer]]
[[def: nonce]]
[[def: predicates]]
[[def: REV_REG_DEF]]
[[def: REV_REG_ENTRY]]
[[def: SCHEMA]]
[[def: subject]]
[[def: verifier]]
```

swcurran:
Better tool for finding all the refs:

[...]

PR #193, opened 2023-12-21
Finalize - pending review - the terminolgy for the specification

swcurran:


swcurran:
Did a pass over the terminology to address the first part of #60. Once I have this done, as I pass through all of the spec docs (for a last pass review), I'll update the "ref"s.

Appreciate a review/approval/comments as I'd like this in place before the next step.

Thanks @TelegramSam @mikelodder7 and anyone willing/able to review.  Perhaps @wip-abramson ?

spec/terminology.md
```diff
@@ -1,137 +1,201 @@
 ## Terminology
 
-[[def: blinding factor]]
-~ A blinding factor is a random 3152 bit selected from the set of integers up to the order of the RSA group, `n`. It is generated by the [[ref:holder]] to blind the their [[ref:link secret]] during credential issuance. The blinding factor can then be removed from the signature the [[ref:issuer]] produces to retrieve a valid signature over the unblinded [[ref:link secret]]. A blinding factor is similarly generated for each non-disclosed attribute during credential presentation, such that a [[ref: holder]] can prove they know these values without revealing them to the [[ref: verifier]].
+[[def: Accumulator, accumulator]]
 
+~ A [cryptographic accumulator] is used space- and time-efficient method of proving a value membership is in a set of values without revealing the individual members of the set. In AnonCreds v1, an accumulator is a core element of the [[ref: verifiable credential]] revocation mechanism.
```
wip-abramson, line 5:
```suggestion
~ A [cryptographic accumulator] is used as a space- and time-efficient method of proving a value membership in a set of values without revealing the individual members of the set. In AnonCreds v1, an accumulator is a core element of the [[ref: verifiable credential]] revocation mechanism.
```

wip-abramson: I would be tempted to reference the type of accumulator used, but maybe this is the wrong place

spec/terminology.md
```diff
@@ -1,137 +1,201 @@
...
+
+[[def: BigNumber, BigNumbers]]
+
+~ A [BigNumber] is an data object which safely allows mathematical operations on numbers of any magnitude. BigNumbers are commonly used in cryptography schemes, including those underlying AnonCreds v1.0.
+
+[BigNumber]: https://docs.ethers.org/v5/api/utils/bignumber/
+
+[[def: Blinded Secret, Blinded secret, blinded secret, unblinded secret, Unblinded secret]]
+
+~ A cryptographic technique where a secret value (a number) is blinded before it is shared such that the sender can later prove knowledge ("unblind") of the secret value. The AnonCreds v1.0 [[ref: link secret]]
```
wip-abramson, line 27:
```suggestion
~ A cryptographic technique where a secret value (a number) is blinded before it is shared such that the sender can later prove knowledge of the secret value. The AnonCreds v1.0 [[ref: link secret]]

This is not what I understand as ublinding. Unblinding is ther removal of the blinding factor from a signature over a blinded value. Such that you get a signature on the unblinded value.

I think just removing as I have suggested is okay.
```

spec/terminology.md
```diff
@@ -1,137 +1,201 @@
...
 
 [[def: Blinded Secrets Correctness Proof]]
-~ A ZKP based proof to show that some blinded secret was produced correctly from an unblinded secret without exposing the secret to anyone.
 
-[[def: credential, credentials]]
+~ A [[ref: ZKP]]-based proof that can be verified to show that a [[ref: blinded secret]] was produced correctly from an [[ref: unblinded secret]] without exposing the secret.
+
+[[def: Blinding Factor, blinding factor]]
+
+~ A blinding factor is a random 3152 bit selected from the set of integers up to the order of the RSA group, `n`. It is generated by the [[ref:holder]] to blind the their [[ref:link secret]] during credential issuance. The blinding factor can then be removed from the signature the [[ref:issuer]] produces to retrieve a valid signature over the [[ref: unblinded link secret]]. A blinding factor is similarly generated for each non-disclosed attribute during credential presentation, such that a [[ref: holder]] can prove they know these values without revealing them to the [[ref: verifier]].
```
wip-abramson, line 36:
```suggestion
~ A blinding factor is a random [[ref: BigNumber]] selected from the set of integers up to the order of the RSA group, `n`. It is generated by the [[ref:holder]] to blind their [[ref:link secret]] during credential issuance. Knowledge of the blinding factor can be used to create a [[ref: Blinded Secrets Correctness Proof]].The blinding factor can be removed from the signature the [[ref:issuer]] produces to retrieve a valid signature over the [[ref: unblinded link secret]]. A blinding factor and associated [[ref: Blinded Secrets Correctness Proof]] are similarly generated for each non-disclosed attribute during credential presentation, such that a [[ref: holder]] can prove they know these values without revealing them to the [[ref: verifier]].
```

spec/terminology.md
```diff
@@ -1,137 +1,201 @@
...
+
+[[def: Call Home, Call home, call home]]
+
+~ Call home is a term used when evaluating the privacy characteristics of [[ref: verifiable credential]] deployments. If a [[ref: holder]] presenting data from a verifiable credential must always contact ("call home to") the [[ref: issuer]], [[ref: holder]] actions are open to the actual, or perception of, tracking of the holder by the issuer. [[ref: Verifiable credential]] schemes that do not make possible the tracking of [[ref: holder]] activities by [[ref: issuers]] are preferred.
```
wip-abramson, line 40: Is it not also if a Verifier must call home to the issuer in order to verify each credential they are presented with?

spec/terminology.md
```diff
@@ -1,137 +1,201 @@
...
+[[def: Credential Definition, Credential Definitions, Public Credential Definition, Private Credential Definition, CredDef, CredDefs, credential definition, CLAIM_DEF, CRED_DEF]]
+
+~ A credential definition (also known as CRED_DEF or CLAIM_DEF) contains data required for [[ref: credential]] issuance (used by the [[ref:issuer]]) as well as [[ref: credential]] validation data (used by the [[ref:holder]] and the [[ref:verifier]]). A credential definition is generated by the [[ref:issuer]] before [[ref:credential]] issuance and consists of two distinct but strongly related parts, the **Public Credential Definition** that is published for everyone to use and **Private Credential Definition** that is held as a secret by the issuer and has the private keys necessary to generate signed [[ref: verifiable credentials]]. A credential definition may be generated such that credentials it produces can be revoked.
```
wip-abramson, line 56: Interesting. I always thought that CRED_DEF refers to only the public part. Here seems to be saying it is both the public and private parts.

spec/terminology.md
```diff
@@ -1,137 +1,201 @@
...
+
+[[def: Credential Key Correctness Proof]]
+
+~ A proof generated during the creation of the [[ref: credential definition]] and included in the [[ref: credential offer]] so that the [[ref:holder]] can verify that the published [[ref: credential definition]] used in the blinding process belongs to the [[ref:issuer]].
```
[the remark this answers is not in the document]
wip-abramson: Yeah see, this makes me thing that credential definition should be public. Then there is some other defined term for the private part.

spec/terminology.md
```diff
@@ -1,137 +1,201 @@
...
 
-~ The **Private Credential Definition** part of a Credential Definition is stored on the [[ref: issuer]]'s side and is an object that contains the [[ref:issuer]]`s private keys for signing the attributes (one key for one attribute) when issuing an AnonCreds [[ref:Credential]]. These private keys are used to create signature proofs for the issued anonymous credentials, which then can be validated in a derived form by a [[ref: verifier]] by using the published public keys of the Public Credential Definition part. A Private Credential Definition never leaves the [[ref: Issuer]]'s domain and is stored securely.
+[[def: Credential Request, credential request, Credential request]]
 
-~ Revokable Verifiable Credentials require (besides) a Credential Definition also a [[ref: REV_REG_DEF]].
+~ A credential request is a request from an [[ref: holder]] towards a [[ref: issuer]] to get a credential issued by the [[ref: issuer]]. The credential request references a preceding [[ref: credential offer]] and defines the claims the [[ref: holder]] wants to get issued. A credential request also includes a [[ref: nonce]] that is later to use the issuance of the credential.
```
wip-abramson, line 68:
```suggestion
~ A credential request is a request from an [[ref: holder]] towards a [[ref: issuer]] to get a credential issued by the [[ref: issuer]]. The credential request references a preceding [[ref: credential offer]] and defines the claims the [[ref: holder]] wants to get issued, including a [[ref: Blinded Secret]] and associated [[ref: Blinded Secrets Correctness Proof]]. A credential request also includes a [[ref: nonce]] that is later to use the issuance of the credential.
```

spec/terminology.md
```diff
@@ -1,137 +1,201 @@
...
-specification.
-
-[[def: holder, holders]]
-~ A holder, also known as an identity holder, is an entity that is in possession of a [[ref: credential]]. In many use cases, the holder is also the identity [[ref: subject]]. A holder can interact with an issuer to obtain anonymous credentials. It can also derive information from anonymous credentials that can be presented to a [[ref: verifier]] to gain access to goods and services.
-
-[[def: issuer, issuers]]
+
+~ A Decentralized Identifier (DID), defined by the [W3C DID Core Specification](https://w3c.github.io/did-core/), is a type of identifier that is useful in enabling verifiable, decentralized digital identity. A DID refers to any subject (e.g., a person, organization, thing, data model, abstract entity, etc.) as determined by the controller of the DID. DIDs are not used in AnonCreds itself but there must be a verifiable identifier (usually a DID) with an enforced relationship between [[ref: schema publishers]] and [[ref: issuers]] and the AnonCreds objects they publish. This is outlined in a note in [this specification section](#anoncreds-setup-data-flow).
```
wip-abramson, line 76: In some earlier instances you defined the links serparetly. Do we want to continue that for the DID core link?

E.g. BigNumber is like this

[BigNumber]: https://docs.ethers.org/v5/api/utils/bignumber/

spec/terminology.md
```diff
@@ -1,137 +1,201 @@
...
+
+[[def: DID Method, DID method, DID Methods, DID methods]]
+
+~ DID methods specify how DID documents are written (registered) and read (resolved) on a given [[ref: verifiable data registry]] implementation, allowing DIDs to be registered and resolved in a wide variety of storage containers. The capabilities required by DID Methods are defined in the [DID Core specification], and the (many) DID Methods are defined in the [DID Methods Registry] repository.
```
wip-abramson, line 80:
```suggestion
~ DID methods specify how DID documents are created and read (resolved) on a given [[ref: verifiable data registry]] implementation, allowing DIDs to be registered and resolved in a wide variety of storage containers. The capabilities required by DID Methods are defined in the [DID Core specification], and the (many) DID Methods are defined in the [DID Methods Registry] repository.
```
```suggestion
~ DID methods specify how DID documents are written (registered) and read (resolved) on a given [[ref: verifiable data registry]] implementation, allowing DIDs to be created and resolved in a wide variety of storage containers. The capabilities required by DID Methods are defined in the [DID Core specification], and the (many) DID Methods are defined in the [DID Methods Registry] repository.
```

A DID document does not always have to be registered to be valid. E.g. did:key. I think created aligns better with DID spec

spec/terminology.md
swcurran: What is the type of accumulator used? Do you know off-hand?  I can check if needed.

swcurran: Updated the text, and updated where the accumulator is used in AnonCreds — revocation.

spec/terminology.md
swcurran: Sounds good — thanks. Done.

spec/terminology.md
swcurran: Nice. Done.

spec/terminology.md
swcurran: It is of far less concern, I think.  That a verifier uses a particular type of ID becomes known, but nothing about the holders they are verifying.  Given the existing interaction between the issuer/holder, that is what enables tracking. A verifier would not be anything but an arbitrary verifier to the issuer.  Its number of verifications would seem to be the most that could be tracked — with lots of techniques to get around that.

spec/terminology.md
swcurran: The “private” data of a Cred Def is just the corresponding data to the public parts, so doesn’t seem to be worth splitting up.

spec/terminology.md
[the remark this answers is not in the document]
swcurran: OK - fair enough. I’ll try a reword...

spec/terminology.md
swcurran: Good catch — aligned to use the one link definition.

spec/terminology.md
swcurran: Updated

spec/terminology.md
swcurran: Reworded.

spec/terminology.md
wip-abramson: It is an RSA based accumulator I believe, but I do not know the exact name offhand

spec/terminology.md
```diff
@@ -1,137 +1,201 @@
...
+[[def: Credential Definition, Credential Definitions, Public Credential Definition, Private Credential Definition, CredDef, CredDefs, credential definition, CLAIM_DEF, CRED_DEF]]
+
+~ A credential definition (also known as CRED_DEF or CLAIM_DEF) contains the public data required for [[ref: credential]] issuance (used by the [[ref:issuer]]) as well as [[ref: credential]] validation data (used by the [[ref:holder]] and the [[ref:verifier]]). A credential definition is generated by the [[ref:issuer]] before [[ref:credential]] issuance and published for anyone (primarily [[ref: holders]] and [[ref: verifiers]]) to use. In generating the public credential definition, related private data is also generated and held as a secret by the issuer. The secret data includes the private keys necessary to generate signed [[ref: verifiable credentials]] that can be presented and verified using the published credential definition. A credential definition can optionally be generated such that the credentials it produces can be revoked.
```
TelegramSam, line 56: This should make clear that one cred def can be used for any number of issuances. It might be inferred that a cred def is required per issuance with the current reading.

spec/terminology.md
```diff
@@ -1,137 +1,201 @@
...
+[[def: Data Minimization, data minimization, Data minimization]]
 
-[[def: Credential Request]]
-~ A credential request is a request from an [[ref: holder]] towards a [[ref: issuer]] to get a credential issued by the [[ref: issuer]]. The credential request references a preceding [[ref: Credential offer]] and defines the claims the [[ref: holder]] wants to get issued. A credential request also includes a [[ref: nonce]].
+~ An attribute of verifiable data sharing schemes that considers privacy of a scheme based on the amount of data shared in a given interaction. Ideally, the minimum amount of data is shared for the business purpose of the interaction. Techniques such as [[ref: selective disclosure]], [[ref: predicates]], and [[ref: unlinkability]], all available in AnonCreds v1.0 support the goal of privacy-preserving, minimal data sharing.
```
TelegramSam, line 72: several times the term `business purpose` is used in definitions. I'm not sure we should include the term `business`, and I think it reads fine without it.

TelegramSam:
a few comments added

spec/terminology.md
swcurran: Updates made — good catch.

spec/terminology.md
swcurran: Removed.

swcurran:
Thanks @TelegramSam — ready for re-review.

TelegramSam (approved):

PR #197, opened 2024-01-24
Clarify revocation description

doowt:
This PR aims to clarify revocation in AnonCreds in terms of the format of generated objects. There was a placeholder suggestion that there should be more detail (possibly in the appendix) on what pairings are and how they work; the goal of this PR is only to explain enough to understand what the objects look like rather than how the revocation actually works.

The description is based on what I could understand by looking at the [anoncreds-clsignatures-rs](https://github.com/hyperledger/anoncreds-clsignatures-rs) repo. This hopefully provides clarity on the sizes of different fields in the JSON objects and what they should look like.


mikelodder7 (approved):

swcurran:
Awesome — thanks for the contribution, sorry about the delay, and thanks for the review.  Good stuff!

